PR #28, opened 2021-09-29
Add crossref generator

chartgerink:
This PR adds the CrossRef generator, based off the [schema for CrossRef XML](https://data.crossref.org/reports/help/schema_doc/4.4.2/index.html) deposits.

This generator creates only the JavaScript object that fits the schema. This will still need to be converted to XML using a module like [`xml-js`](https://github.com/nashwaan/xml-js) before submitting to CrossRef.

⚠️ This is a work in progress ⚠️ 

## Todo

- [x] Figure out what type to use for depositing research module (meeting with CrossRef on Thursday 30 sep about this)
- [x] Generate body function
- [x] Test upload.


chartgerink:
After meeting with CrossRef on what type - it seems like book chapters are the best fitting our discoverability needs and metadata needs. Updating approach now...

chartgerink:
Finally in a good state!

Four automated deposits in a row went perfectly good!

Screenshot 2021-12-30 at 21 56 31

Most important to note is that the environment variable for the login requires the role to be included, `email/role` which I stumbled over.

PR #32, opened 2021-10-05
Remove old session data periodically

chartgerink:
This is WIP on removing session data periodically to clean out the database (each unique user + device combination gets a session). 

At the  moment this is blocked because BlitzJS is having some issues with [the recipe installer](https://github.com/blitz-js/blitz/issues/2793).

## Todo

- [ ] `blitz install quirrel` (when it works again)
- [ ] [Create CronJob](https://blitzjs.com/docs/background-processing-with-quirrel#removing-old-data)
- [ ] Verify functionality

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,14 +34,6 @@
         "source-map": "^0.5.0"
       },
       "dependencies": {
-        "json5": {
-          "version": "2.2.0",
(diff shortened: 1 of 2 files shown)


chartgerink:
Superceded by #140

PR #47, opened 2021-10-15
Change navigation bar

chartgerink:
This PR updates the navigation bar according to the ongoing designs. This is in the grayboxing stage and is primarily an implementation of the structure, not the details.

The figma files can be referenced [here](https://www.figma.com/file/XH0yXo3KkmBZVCeDB97OoS/researchequals.com?node-id=25%3A256).


chartgerink:
Now includes a workspace switcher 😊 (although no way to create additional workspaces just yet 😉)

https://user-images.githubusercontent.com/2946344/138069177-e450d14f-05ed-49f4-bc35-1726ce785321.mov

PR #159, opened 2021-12-22
Update styling of nav

chartgerink:
This PR adds a bunch of styling work of today. 

Started out intending to do the navigation, but worked rolled over into the profile page.

Screenshot 2021-12-22 at 17 48 50
Screenshot 2021-12-22 at 17 51 39

Still struggling with getting the settings modal to pop up when the dropdown menu is activated. Or well, have it work properly and not close as soon as I try to do anything in the settings modal. To be continued...


psobrakseaton:
If opening the modal from the dropdown is proving to be too troublesome, we could also add a gear icon to the menu bar and have it open from there.

PR #162, opened 2021-12-23
`[handle]` styling

chartgerink:
This PR adds a bunch of styling to the `[handle]` pages.

Screenshot 2021-12-23 at 14 38 36
Screenshot 2021-12-23 at 15 01 58

Still missing is the popover panel for the people they're following, but that'll be added still.


chartgerink:
Now with the following panel (vertical centering of the text took me a moment...)

Screenshot 2021-12-23 at 19 20 56
Screenshot 2021-12-23 at 19 21 45

PR #161, opened 2021-12-22
Add stripe checkout

chartgerink:
This PR integrates Stripe checkout for publishing of modules. This is a WIP.

My understanding of how [Stripe Checkout](https://stripe.com/docs/payments/checkout) works is the following:

Untitled Diagram drawio

- `api/checkout_sessions` creates the Stripe Checkout session to fulfill the payment.
- When Checkout completed, the user gets returned to whatever success link on R=
- When Stripe completes the payment, it can send information to a webhook (e.g., `api/stripe_webhook` in this diagram)
- `api/stripe_webhook` then completes the publication of the module on the relevant license

Note that payment completes instantaneously for some payment options, and is [delayed for others](https://stripe.com/docs/payments/checkout/fulfill-orders#delayed-notification).

## Todo

- [x] Add session creation
- [x] Add checkout session request handling
- [x] Add webhook for payment completion
- [ ] ~~Update module database for pending publications (payment ongoing, webhook not received)~~
- [x] Add `price_id` to licenses
- [ ] ~~Add `customer_id` to User~~
- [x] Figure out tax id collection
- [ ] ~~Retrieve customer_id upon signup for user based on email?~~

## Resources

- [Stripe Checkout Quickstart](https://stripe.com/docs/checkout/quickstart)
- [Handle delayed notifications](https://stripe.com/docs/payments/checkout/fulfill-orders#delayed-notification)
- [Webhooks](https://stripe.com/docs/webhooks)

PR #166, opened 2021-12-25
Update drafts

chartgerink:
This PR updates the drafts page to the new designs. Work in progress, pushing to not lose any local work.

This includes a complete rework of the "inbox" style display - there were some issues using HeadlessUI's disclosure on iOS - the display of the buttons was funky. 

Screenshot 2021-12-25 at 12 07 50

## Todo

- [x] Edit module
- [x] Publish module (see also #28 and #161 to complete those)
~~- [ ] Follows from linking~~
- [x] Manage authors


chartgerink:
Now with improved author management 
Screenshot 2021-12-28 at 14 39 09

chartgerink:
Editing is in a semi better place - now at least in the same location as the regular metadata but still requires styling.

Screenshot 2021-12-28 at 14 51 16

psobrakseaton:
> Editing is in a semi better place - now at least in the same location as the regular metadata but still requires styling.
> 
> Screenshot 2021-12-28 at 14 51 16

This is really good. I have had some ideas over Christmas as to make this smoothed out, but this is a step in the right direction.

chartgerink:
Now with publishing action (still requires #161 and #28 to be finalized).

Screenshot 2021-12-28 at 15 30 52

The modal that comes from it is unstyled at the moment.

Screenshot 2021-12-28 at 15 33 23

chartgerink:
Note: Depends on #167 before merging.

PR #168, opened 2021-12-29
Published module display

chartgerink:
This PR adds the published module display. This was rather straightforward because I implemented all relevanted elements for the invitations screen.

Screenshot 2021-12-29 at 11 52 01

Note that the actions for published modules are not yet defined and hence not implemented.


chartgerink:
Depends on #169 being merged into this PR.

PR #169, opened 2021-12-29
Upgrade dashboard

chartgerink:
This PR upgrades the dashboard display.

Screenshot 2021-12-29 at 14 07 07
Screenshot 2021-12-29 at 14 07 14

Most of note change is  that email verification is now an onboarding quest as well.


chartgerink:
Depends on #170 being merged into this first.

PR #176, opened 2022-01-03
Add events

chartgerink:
This PR is for adding the events to the website that are needed to create the (minimal) conversion funnels of interest.

Using Splitbee, we'll track some core information about the usage of the website in the aggregate.

The core of this data-driven approach to R= at launch are the following funnels

1. `/` --> `/signup` --> Sign up
2. Sign up --> Publish first module --> Publish second module

We're primarily interested in these two funnels at first.

## Todo

- [x] Sign up
- [x] Publish first module
- [x] Publish second module


chartgerink:
This now sends an event to Splitbee whenever somebody hits the publish button. It does not at this point make a difference whether payment is completed, which may distort the data somewhat.

`Publish module ${x}` is the event name.

PR #191, opened 2022-01-05
Add reference feature

chartgerink:
This PR introduces the final feature before launch: Reference lists.

References are managed in the same table as modules on the back-end, which will allow us to do some cool stuff in the future. For now, you can add a structured reference list to your modules. If we don't yet have the DOI in our database, people can input a DOI and add it to the database directly (only for CrossRef now).

https://user-images.githubusercontent.com/2946344/148239410-56654d4c-55c0-4c19-97a9-db49a09d6fc0.mov

As a side benefit: You can also link your modules to anything published elsewhere. This means that the parent of a module can be virtually any research output!

There will be some additional required QA on this for design (@psobrakseaton). This feature manifested itself as a necessity for launch during development, so it is a bit unexpected in that sense.

### DIFF
--- a/app/api/stripe_webhook.ts
+++ b/app/api/stripe_webhook.ts
@@ -63,6 +63,15 @@ const webhook = async (req: BlitzApiRequest, res: BlitzApiResponse) => {
               workspace: true,
             },
           },
+          references: {
+            include: {
+              authors: {
+                include: {
+                  workspace: true,
+                },
+              },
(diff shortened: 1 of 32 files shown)


chartgerink:
This looks to work well this time around :-)

This does still require further design styling to occur during QA, because font sizes are out of sync with the rest (the rest is too small right now).

PR #197, opened 2022-01-06
Add Apple Pay domain verification

### LINKED ISSUE #172: Apple Pay integration
Upon discovery in #161, [Apple Pay](https://support.stripe.com/questions/enable-apple-pay-on-your-stripe-account) requires additional work to be integrated into the Stripe Checkout pattern.

This issue helps track the work on that. The hurdle that needs to be cleared (AFAIK) how to add the merchant file in a way that it doesn't have to be tracked in the repo (considering it is like a secret key but needs to be a file it seems).

chartgerink:
This PR adds Apple Pay domain verification.

Has to be merged straight into main because I have to finish the setup process now.

Fixes #172.


chartgerink:
Ah damn I branched this from staging so it's picking up all those changes. Hold on... (closing and opening a better one)

PR #243, opened 2022-01-17
Add code of conduct

chartgerink:
This PR adds the code of conduct from our Notion version. This PR supercedes #212.

https://user-images.githubusercontent.com/2946344/149939433-4e0abf4e-0405-4962-b34d-db2f55664125.mov

We now also add a quick report button that puts all relevant information in the body of the email.

https://user-images.githubusercontent.com/2946344/149939459-e42d7017-b1ce-409b-b6e5-8a70a1779aca.mov

Technical Note: We can use `dangerouslySetInnerHtml` here because we control the markdown. In other situations that will not be responsible as people can inject html we don't control. Future changes should not be allowed to include raw html in the markdown code.


chartgerink:
Getting some render issue on the phone that needs to be fixed first image

chartgerink:
Thanks @InquisitiveVi! That is helpful. I will make the update accordingly.

The `coc@libscie.org` automatically goes to all members of the committee, so no extra work involved in that sense. 😊 

@all-contributors add @InquisitiveVi for a11y

PR #254, opened 2022-01-21
Add TDM meta tags

chartgerink:
This PR adds a machine readable meta tag to help inform TDM permissions/restrictions. 

It implements both the `robots.txt` option with limited snippets and the unofficial draft of the W3c working group. In that sense, this is now also used to explicitly *allow* the use of TDM, not only to restrict it. The condition for restriction is when a module is published under "All rights reserved"

This is a requirement based on the [DIRECTIVE (EU) 2019/790](https://eur-lex.europa.eu/legal-content/EN/TXT/HTML/?uri=CELEX:32019L0790&from=EN).


chartgerink:
@all-contributors add @senficon for ideas and research

PR #309, opened 2022-02-08
Add linked data JSON to headChildren

### LINKED ISSUE #294: pages of research output (with an DOI) should have schema.org annotation to boost SEO
# Context
Over the past couple of years I have worked with the [Bioschemas](https://bioschemas.org/) team to add semantic annotation to multiple website. It is an extension in the life sciences of [schema.org](https://schema.org/) which also has annotation for "types" like Computational Tool, Data/Data Set, etc. The annotation is normally autogenerated from content already on the page. 

# Expected Outcome
Published Steps have schema.org/Bioschemas JSON-LD embedded in their pages, to be indexed by all major search engines.

I am happy to support this and will try to find time to make a draft patch at some point.

# Example

This examples comes from the schema.org website and shows the HTML element one would embed somewhere in the HTML page for a `CreativeWork`, possible the simplest solution for any "Step":

```xml
<script type="application/ld+json">
{
  "@context": "https://schema.org",
  "@type": "CreativeWork",
  "author": "Sony",
  "contentRating": "Mature",
  "image": "videogame.jpg",
  "name": "Resistance 3: Fall of Man"
}
</script>
```

Many fields are optional and a rich overview is provided at https://schema.org/CreativeWork

chartgerink:
This PR adds the linked data JSON to the head for published modules.

There is ample room to include more information, but here we choose to implement a first version now rather than more extensive functionality much later. We can still add layers to this as needs arise, so feel free to share those as they come up.

Fixes #294.﻿

### DIFF
--- a/app/pages/modules/[suffix].tsx
+++ b/app/pages/modules/[suffix].tsx
@@ -1,8 +1,8 @@
 import { Prisma } from "prisma"
 import { Link, NotFoundError, Routes, useMutation, useQuery, useRouter, useSession } from "blitz"
-import { AddAlt32, Undo32, NextFilled32, PreviousFilled32 } from "@carbon/icons-react"
+import { AddAlt32, NextFilled32, PreviousFilled32 } from "@carbon/icons-react"
 import Xarrows from "react-xarrows"
-import { useEffect, useState } from "react"
+import { useState } from "react"
 
 import Layout from "../../core/layouts/Layout"
 import db from "db"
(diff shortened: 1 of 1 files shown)


egonw:
@chartgerink, you need to escape the quotes in titles and descriptions. Here's the validation report of one page:

https://www.researchequals.com/modules/r5rj-vthh

Not the error, due to the unescaped quote.

PR #322, opened 2022-02-11
Improve JSON-LD data

### LINKED ISSUE #313: escaping of the JSON-LD is wrong, in two ways
The following code is a good start and the JSON-LD content generally looks great:

There is two current issues that should be easy to fix (tho, not sure how to do it):

1. right now the full JSON is escaped in the HTML, but the quotes in the HTML should overall really be quotes and not HTML escape coded. I think the overall escaping is triggered by the backtick at the end on this line:

https://github.com/libscie/ResearchEquals.com/blob/fcbe4f7809e0a2c95fee74683d889d1e856687e0/app/pages/modules/%5Bsuffix%5D.tsx#L351

2. the quotes *inside* field values should be escaped, however, in the title and definition, where quotes can be part in the these field values. This is for example on this line:

https://github.com/libscie/ResearchEquals.com/blob/main/app/pages/modules/%5Bsuffix%5D.tsx#L357

In pseudocode, this should look something like this:

```
"description": "${module.description.replace("\"","&quot;")}",
```

chartgerink:
This PR removes the escaping of the JSON-LD data.

- Add react-helmet
- Improve JSON-LD data

Fixes #313.

### DIFF
--- a/app/pages/modules/[suffix].tsx
+++ b/app/pages/modules/[suffix].tsx
@@ -340,6 +340,18 @@ const Module = ({ module, mainFile, supportingRaw }) => {
         <ParentPanel openObject={previousOpen} openFunction={setPreviousOpen} module={module} />
         <ChildPanel openObject={leadsToOpen} openFunction={setLeadsToOpen} module={module} />
       </article>
+      <Helmet>
+        <script className="structured-data-list" type="application/ld+json">
+          {JSON.stringify({
+            "@context": "https://schema.org",
+            identifier: `${module.prefix}/${module.suffix}`,
+            "@type": "CreativeWork",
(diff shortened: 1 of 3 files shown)


chartgerink:
Looks like this validation works 😊 

@egonw to prevent from a premature merge from my side (again) - is there anything else that's top of mind for you if the validator passes without errors/warnings?

Screenshot 2022-02-11 at 13 03 09

egonw:
No, I would suggest to merge this in. Next step is to map the various module types to more specific schema.org types.

chartgerink:
Thanks @egonw! I will merge this and take a looksie over in your much appreciated PR 😊

PR #328, opened 2022-02-14
Test session feedback (nr. 12)

chartgerink:
This PR includes feedback from test session number 12.

Nothing groundbreaking - general fixes. Most important is that a single author was apparently able to publish a module without entering their author names - this is now fixed.


chartgerink:
@all-contributors add @SarahanneMField for userTesting

PR #327, opened 2022-02-13
Added (initial) schema.org mappings of module types

egonw:
Which can be used to update/extend https://github.com/libscie/ResearchEquals.com/pull/322

Issue #151, opened 2021-12-16
Improve seed script for pricing updates


chartgerink:
The seed script `db/seed.ts` currently (as of #150) only creates the licenses and doesn't update based on changes made to the file.  If it would update the database if there are changes, that would be a helpful way to manage pricing changes to licenses in the future.

This can be done by replacing the `createMany` statement with an `upsert` command for the licenses. This will make the seed script a bit longer (one `upsert` statement per license instead of one `createMany` statement) - it will also help improve pricing management.

PR #331, opened 2022-02-15
Add `schema` to `ModuleType`

chartgerink:
This PR builds on #327 - I cannot edit the fork so have to create a new PR. I add the extension to the `ModuleType` data schema, to include the `schema` and the database migration for production.

@egonw's original contributions are still reflected in the commits.

The only downside: Because the module types are already in the database, the seed script will skip these items. This means that the modules that are not `CreativeWork` need to be updated manually in the production database. We could revisit this at a later time when #151 has been implemented - to ensure that the database gets updated if anything has changed.


chartgerink:
@all-contributors add @egonw for code

PR #334, opened 2022-02-15
Match any DOI in a string and use it as a query

### LINKED ISSUE #325: Cannot add references by DOI formatted as an URL
I was trying to add references to a module, but I couldn't do so by a DOI in an URL format (e.g., I cannot use `https://doi.org/10.1037/1089-2680.10.3.229`, but I can use `10.1037/1089-2680.10.3.229`)

image

It would be nice to add references using either string.

nsunami:
This PR expands the matching of a DOI string so that any string containing a DOI will be matched and be added as a reference. (fix #325)

I was not able to run this on local so the changes are not verified. (sorry!) But hope it captures the idea.

### DIFF
--- a/app/modules/components/ModuleEdit.tsx
+++ b/app/modules/components/ModuleEdit.tsx
@@ -269,15 +269,21 @@ const ModuleEdit = ({
                         return <SearchResultModule item={item} />
                       },
                       noResults() {
+                        const matchedQuery = query.match(/10.\d{4,9}\/[-._;()/:A-Z0-9]+$/i)
+                        const matchedDoi = matchedQuery.slice(-1)
                         return (
                           <>
                             {/* https://www.crossref.org/blog/dois-and-matching-regular-expressions/ */}
-                            {query.match(/10.\d{4,9}\/[-._;()/:A-Z0-9]+$/i) ? (
+                            {matchedDoi ? (
                               <>
                                 <button
                                   className="text-sm font-normal leading-4 text-gray-900 dark:text-gray-200"
                                   onClick={async () => {
-                                    toast.promise(createReferenceMutation({ doi: query.match(/10.\d{4,9}\/[-._;()/:A-Z0-9]+$/i) }), {
+                                    toast.promise(
+                                      createReferenceMutation({
+                                        doi: matchedDoi,
+                                      }),
+                                      {
                                       loading: "Searching...",
                                       success: (data) => {
                                         toast.promise(
@@ -299,10 +305,11 @@ const ModuleEdit = ({
                                         return "Record added to database"
                                       },
                                       error: "Could not add record.",
-                                    })
+                                      }
+                                    )
                                   }}
                                 >
-                                  Click here to add {query.match(/10.\d{4,9}\/[-._;()/:A-Z0-9]+$/i)} to ResearchEquals database
+                                  Click here to add {matchedDoi} to ResearchEquals database
                                 </button>
                               </>
                             ) : (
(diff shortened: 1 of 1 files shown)


chartgerink:
Thanks for the PR @nsunami 🙏 I just tried this out locally and the code itself looks good.

```js
> let x = 'https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0127502'
> x.match(/10.\d{4,9}\/[-._;()/:A-Z0-9]+$/i)
[
  '10.1371/journal.pone.0127502',
  index: 45,
  input: 'https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0127502',
  groups: undefined
]
```

I tried testing it a bit in the development build and stumbled upon the following:

```mermaid
flowchart TD
    http://doi.org/10.1629/uksg.558 --> https://insights.uksg.org/articles/10.1629/uksg.558/ 
```

The original link for the DOI works without a hitch. 

The resolved URL for the DOI, in this instance, has a trailing slash that will be matched (`10.1629/uksg.558/`). This leads to an error for the metadata retrieval because this is an invalid DOI (see `http://api.crossref.org/works/10.1629/uksg.558/`).

We could remove the trailing slash, but that is ad hoc. It makes me wonder about other URLs and their (potential) exceptions. Even worse, we may have conflicting exceptions in URLs across publishers, for whatever odd reasons 😉 

I agree with your original issue #325 that adding a DOI link option is worthwhile (e.g., `https://doi.org/10.1629/uksg.558`). I wonder whether including publisher links that may include a DOI is worth it (e.g., `https://insights.uksg.org/articles/10.1629/uksg.558/`). The latter requires rigorous quality assurance and test cases, as we start seeing here.

What do you think @nsunami?

In any case, @all-contributors add @nsunami for code 💜

chartgerink:
I am discussing this with @nsunami (synchronously) at the moment and discussing what should be matched: 

```
https://doi.org/10.1629/uksg.558
doi.org/10.1629/uksg.558
http://doi.org/10.1629/uksg.558
https://dx.doi.org/10.1629/uksg.558
http://dx.doi.org/10.1629/uksg.558
10.1629/uksg.558
```

It seems like we're leaning towards not including non-DOI links, nor edge case DOIs that aren't covered already:

```
https://insights.uksg.org/articles/10.1629/uksg.558/
10.1002/1099-0690(200204)2002:8<1397::AID-EJOC1397>3.0.CO;2-6
```

nsunami:
I came up with this expression. It may be too verbose though--but it's matching the examples above.

`^(?:(?:(?:http|https):\/\/|)(?:dx.doi.org|doi.org)\/)?(10.\d{4,9}\/[-._;()\/:A-Z0-9]+$)`

chartgerink:
Nice @nsunami 🔥 Regular expressions are tough, and something is better than nothing!

I tried to simplify it a tiny bit based on yours:
```
^((http)s?:\/\/)?((dx\.)?doi\.org\/)?(10.\d{4,9}\/[-._;()\/:A-Z0-9]+$)
```

This should not be used to extract the DOI itself indeed, that match should remain `10.\d{4,9}\/[-._;()/:A-Z0-9]+$`

app/modules/components/ModuleEdit.tsx
```diff
@@ -269,15 +269,23 @@ const ModuleEdit = ({
                         return <SearchResultModule item={item} />
                       },
                       noResults() {
+                        const matchedQuery = query.match(
+                          /^((http)s?:\/\/)?((dx\.)?doi\.org\/)?(10.\d{4,9}\/[-._;()\/:A-Z0-9]+$)/i
+                        )
+                        const matchedDoi = matchedQuery.slice(-1)
```
nsunami, line 275: I ended up assuming that the last element of the matched array is the DOI. I think a better approach would be to implement a look-behind in the regex, but I could not figure out the expression 😢

chartgerink: Did you test that assumption? Either in a test version of the app or locally in your node cli.

Maybe the URLs from the Regex are a good start.

To consider merging this we'd need to be confident that this works 😊 I did a quick test run in my cli for one DOI and that one worked.

nsunami: I'm yet to log in to the app locally (still setting up and getting close), but I tested the logic on [Code Pen](https://codepen.io/nsunami/pen/YzEEKWM?editors=0011). It looks like it's working there. I'm happy to test them on my local once I get this working, but I need a bit more time.

chartgerink: That's perfect, thanks @nsunami 😊 

I will do a local run, and the code pen looks promising.

nsunami: Great! I hope this works on the app!

chartgerink:
@nsunami I tried testing it in the app but this doesn't work when there is **no** DOI.

When I for example use `https://oxfamilibrary.openrepository.com/handle/10546/621341` it returns the following error

Screenshot 2022-03-02 at 11 23 42

I updated the code as follows:

* Removed the `matchedDoi` variable
* Logic based on `matchedQuery` only
* If there is a match, use the slice command where it previously had `matchedDoi`

This only has the odd result that when the DOI is already in the database (e.g., `10.53962/0000`) that the URL cannot be added (e.g., `https://doi.org/10.53962/0000`) - the search does not filter out the already indexed items at the moment.

What do you think?

PR #351, opened 2022-03-02
[Cont.] Match any DOI in a string and use it as a query

chartgerink:
This PR is a continuation of #334.

- Match any DOI in a string and use it as a query
- Update the regex pattern and postprocessing
- Update the regex
- Fix nonmatch error

## Last comment in #334

@nsunami I tried testing it in the app but this doesn't work when there is **no** DOI.

When I for example use `https://oxfamilibrary.openrepository.com/handle/10546/621341` it returns the following error

Screenshot 2022-03-02 at 11 23 42

I updated the code as follows:

* Removed the `matchedDoi` variable
* Logic based on `matchedQuery` only
* If there is a match, use the slice command where it previously had `matchedDoi`

This only has the odd result that when the DOI is already in the database (e.g., `10.53962/0000`) that the URL cannot be added (e.g., `https://doi.org/10.53962/0000`) - the search does not filter out the already indexed items at the moment.

What do you think?

### DIFF
--- a/app/modules/components/ModuleEdit.tsx
+++ b/app/modules/components/ModuleEdit.tsx
@@ -269,15 +269,21 @@ const ModuleEdit = ({
                         return <SearchResultModule item={item} />
                       },
                       noResults() {
+                        const matchedQuery = query.match(/10.\d{4,9}\/[-._;()/:A-Z0-9]+$/i)
+                        const matchedDoi = matchedQuery.slice(-1)
                         return (
                           <>
                             {/* https://www.crossref.org/blog/dois-and-matching-regular-expressions/ */}
-                            {query.match(/10.\d{4,9}\/[-._;()/:A-Z0-9]+$/i) ? (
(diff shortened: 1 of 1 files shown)


nsunami:
I see! Your solution sounds good to me.

From a slightly different angle, I wonder if a better way is to use the regex look behind to avoid slicing. I got the look behind working on regex ([regex101](https://regex101.com/r/ceBB8V/1); replacing `?` to `.` since look behind does not work in variable length) But, I could not make it work on JS ([Codepen](https://codepen.io/nsunami/pen/MWOZxgr?editors=0012))

chartgerink:
This is now fixed - I tested it with a DOI that failed in the beginning because of the trailing slash too.

@nsunami I think this is ready to merge after the test deployment succeeds! Thank you so much for your contribution and patience in getting this included. 🙏

nsunami:
Of course! Glad to be part of it 😄

Issue #354, opened 2022-03-03
Add PDF preview for main file


chartgerink:
Based on the discussion in #346 and the results from #352, a PDF previewer in the browser is desired as it is the most common file extension used at the moment.

![image](https://user-images.githubusercontent.com/2946344/156567188-46e340f0-c78d-423f-b09f-6063c3c44db6.png)

This issue tracks the implementation of this desired feature. It is slated for [March release](https://github.com/libscie/ResearchEquals.com/milestone/2).

## Resources

- [`pdf.js`](https://github.com/mozilla/pdf.js) (some help docs [here](https://stackoverflow.com/questions/55371402/how-to-reference-pdf-js-library-in-react))
- [React PDF viewer](https://react-pdf-viewer.dev/)
- [`react-pdf`](https://projects.wojtekmaj.pl/react-pdf/)

chartgerink:
Doing a bit more research and [this Medium post](https://blog.bitsrc.io/top-react-pdf-libraries-in-2021-d0cfbeccdbf8) has a good compilation of options. Summarized (some redundant with the OP):

1. [`react-pdf/renderer`](https://www.npmjs.com/package/@react-pdf/renderer) - primarily when creating PDFs using React (good to note!)
2. [`react-pdf`](https://www.npmjs.com/package/react-pdf)
3. [`@phuocng/react-pdf-viewer`](https://www.npmjs.com/package/@phuocng/react-pdf-viewer)
4. [`react-file-viewer`](https://github.com/plangrid/react-file-viewer) - covers more than just PDFs ([here's an example](https://codesandbox.io/s/lnyqr))
5. [`@mikecousins/react-pdf`](https://www.npmjs.com/package/@mikecousins/react-pdf)

I especially like option 4 because it would automatically allow us to also preview `docx`, `csv`, and audio/video files without much hassle. This would reduce the amount of dependencies (potentially) needed too.

PR #358, opened 2022-03-21
Add PDF preview

chartgerink:
This PR adds a PDF preview for published modules.

This chooses option 3 from #354 -  [`@react-pdf-viewer/core`](https://react-pdf-viewer.dev/). Before merging, [the license needs to be bought](https://phuocng.gumroad.com/l/react-pdf-viewer?variant=Organization%20license) in order to permit usage. I will do that after the tests complete.

I chose this option because it seemed to provide the most scalable solution into the future. It provides accessibility features and seems to be in active development overall. Also, option 4 fell by the wayside because testing the previewer for other file formats seemed like it would create scope creep for this specific issue. I'd rather revisit previewing data separately and ensure everything works then instead of semi doing it now.

This is a bare bones implementation and does not go beyond displaying the PDF. There are [plugins](https://react-pdf-viewer.dev/plugins/) we can implement in the future.

## Screenshot

Screenshot 2022-03-21 at 17 30 16

## Todo 

- [ ] Buy license


chartgerink:
All works - license bought - merging in 3-2-1 🚀

PR #369, opened 2022-04-21
Redirect to home when no token is found on the reset-password page

### LINKED ISSUE #368: The reset password page is exposed
I noticed that the page to reset the password is exposed. 

https://www.researchequals.com/reset-password

I expected that accessing this URL without logging in (and without initiating the reset password process) would re-direct to the login page.

Maybe a solution would be to verify that the token is there, and if not redirect to the login page?

nsunami:
This PR adds a redirect to the reset-password page. When a user without a token visits the page, they will be redirected to the home page. 

The behavior is handled on the client side. (via `useEffect`)

Fixes #368.

### DIFF
--- a/app/auth/pages/reset-password.tsx
+++ b/app/auth/pages/reset-password.tsx
@@ -1,14 +1,29 @@
-import { BlitzPage, useRouterQuery, Link, useMutation, Routes, validateZodSchema } from "blitz"
+import {
+  BlitzPage,
+  useRouterQuery,
+  Link,
+  useMutation,
+  Routes,
+  validateZodSchema,
+  useRouter,
+} from "blitz"
 import Layout from "app/core/layouts/Layout"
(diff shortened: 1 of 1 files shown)


chartgerink:
Thanks @nsunami for opening the issue and contributing a PR. You're a legend!

I forgot Heroku's auto-deploy doesn't work at the moment (or needs to be re-enabled) so I'll test this first thing next week!

While I'm here - how would you feel about becoming a recognized maintainer? We can chat about it sometime if you'd like - your contributions are always a pleasure 🤗

nsunami:
Happy to help—let's chat!

chartgerink (approved):
This looks great @nsunami!

I verified that both the absence of a token and the presence of a token lead to the expected results. 

Sorry about the Heroku deployment issues - fingers crossed they resolve that soon as it provides additional confidence in deployment. 

Will merge momentarily! 🚀 You rock!

PR #373, opened 2022-04-26
Add Crisp chat

### LINKED ISSUE #289: GDPR Compliant Chat window
We previously intended to use Chatra for our chat, but this is not GDPR compliant - the Standard Contractual Clauses provided by Chatra are outdated and hence are not up to our standards. This resulted in the removal of the Chatra integration in #288.

Need: New chat integration
Requirements: 
- [ ] Up to date Standard Contractual Clauses
- [ ] GDPR Compliant
- [ ] Good mobile app for responding on the go

If anybody has some good suggestions, feel free to drop them in this issue!

## Options explored

- ~~Chatra~~
- [Crisp](https://crisp.chat/en/)

chartgerink:
This PR adds the Crisp chat to the platform.

I was able to add the chat only upon page reload after consenting to the cookie. Re-rendering the `<head>` elements using hooks was something I struggled with a bit but didn't get to work in the way I wanted. 

Additional updates:
* Privacy policy (permitted to be changed like this, terms would require a legal notice to the users I think four weeks in advance as it changes the contract)
* Cookie consent banner

Fixes #289.

### DIFF
--- a/app/core/components/Footer.tsx
+++ b/app/core/components/Footer.tsx
@@ -6,7 +6,7 @@ const Footer = () => {
   const date = new Date()
 
   return (
-    <footer className="w-full bg-white py-4 dark:bg-gray-900 sm:p-0">
+    <footer className="w-full bg-white pt-4 pb-24  dark:bg-gray-900 sm:p-0">
       <div className="mx-4 pt-8 text-black dark:text-white sm:flex sm:max-w-7xl xl:mx-auto ">
         <div className="mb-0 mt-0 w-10/12 text-sm sm:mb-28 sm:w-full">
           <h3 className="text-base font-bold">ResearchEquals</h3>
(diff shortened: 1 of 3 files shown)


nsunami:
I pushed a commit here to resolve the problem of the icon not showing up without a reload. I used a state to track the consent. Would that work? (feel free to roll back if not)

nsunami (approved):

chartgerink:
Ah perfect @nsunami! 

Did you test this before pushing the changes? It does work - and primarily wasn't sure whether you think or knew this worked 😊 

I agree suggesting changes might be a good route in the future - that way the discussion can be localized and the changes can be tracked. I also don't know whether this is possible in VS Code - I would personally do this in GitHub directly.

nsunami:
Yes, I tested this locally. I will try using suggestions next time! 👍

PR #375, opened 2022-04-26
Remove search bar duplication

### LINKED ISSUE #295: Changing between breakpoints duplicates search input
If you change between the breakpoints below and above 680px, the search input can be duplicated.

I don't think it's a major issue since I only seem to being able to reproduce it when changing between the breakpoints pretty fast. Nevertheless, it looks like it could be a memory leak or something that causes this?

https://user-images.githubusercontent.com/770632/151990379-eb7f1496-b6fe-44be-8fb9-642b32b99c55.mp4

chartgerink:
This PR forces the CSS to only display the first of the algolia search bars to remove the issue reported in #295.

Fixes #295.

https://user-images.githubusercontent.com/2946344/165318757-ef8b2e84-240f-49d3-8320-9cde653790dc.mov

### DIFF
--- a/app/core/styles/algolia.css
+++ b/app/core/styles/algolia.css
@@ -85,6 +85,9 @@
 .aa-Autocomplete {
   height: 100%;
 }
+.aa-Autocomplete:not(:first-child) {
+  display: none;
+}
 .aa-DetachedFormContainer *,
 .aa-Panel * {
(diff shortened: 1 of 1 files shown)


nsunami (approved):
Verified that the search bar does not duplicate ✅

PR #376, opened 2022-04-27
Upgrade @react-pdf-viewer/core

chartgerink:
Simple upgrade after all - took longer to create the issue and track the work than to do it... :')


nsunami (approved):
Checked that the app can render a pdf on my local!

PR #379, opened 2022-05-09
Update QUIRREL domain

chartgerink:
I keep receiving 401 errors for quirrel, and I am not sure where it comes from. I know that we switched to the main domain being `www.` so maybe this fixes it?

image

This is a non-breaking change given that it's already broken - simply making a PR for documentation.


nsunami:
I agree with the process! 👍

PR #380, opened 2022-05-10
Upgrade `react-pdf-viewer`

chartgerink:
Another  upgrade. :-) Nothing too special - does make me think there's a lot of outdated dependencies that need to be upgraded and functionality ensured.


nsunami (approved):
Confirmed on local that the app renders a PDF normally 🙌

PR #382, opened 2022-05-10
Upgrade `*tailwindcss*`

chartgerink:
This PR upgrades the `*tailwindcss*` dependencies.

These did not break anything upon initial test and seem to be either [SemVer](https://semver.org/) minor or patch upgrades.


nsunami (approved):
I was able to run the app on my local. Did not notice any UI breaking on my end. 👍

PR #384, opened 2022-05-16
Cascade-delete the token when deleting a user

### LINKED ISSUE #372: Deleting an account fails when a token exists for a password reset
Deleting an account fails when there is an existing token for resetting a password.

## Steps to reproduce
1. Create an account
2. Log out 
3. Do "forgot password" 
4. Log in to the account (without actually resetting a password)
5. Delete account --> 500 error 

image

```
Error: Foreign key constraint failed on the field: `Token_userId_fkey (index)`
```

```
Uncaught (in promise) Error: Foreign key constraint failed on the field: `Token_userId_fkey (index)`
```

I think the solution is to set up a cascading delete for the Prisma schema. (e.g., https://github.com/nsunami/postreview-app/commit/c82d12222fb65dbb4c8f967bb51df7bb09e4695c)

Maybe a small bug, but it may be a good practice to cascade-delete the existing tokens.

nsunami:
This PR deletes any existing token with a user when they delete their account (via adding cascading deletes to the Prisma DB schema). On my local, I checked that the 500 error does not occur with the procedure reported in #372. Fixes #372.

### DIFF
--- a/db/migrations/20220516210705_cascade_delete_token/migration.sql
+++ b/db/migrations/20220516210705_cascade_delete_token/migration.sql
@@ -1,0 +1,5 @@
+-- DropForeignKey
+ALTER TABLE "Token" DROP CONSTRAINT "Token_userId_fkey";
+
+-- AddForeignKey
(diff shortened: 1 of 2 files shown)


chartgerink (approved):
Perfect, thanks @nsunami!

Tested it locally and it behaves as expected. Can confirm it resolves #372 🙏

PR #385, opened 2022-05-17
Upgrade `@carbon/icons-react`

chartgerink:
This PR upgrades `@carbon/icons-react` to `v11`. 

This was a bit of a pain to upgrade, because it introduced minor but breaking changes that were not documented on the [repository itself](https://github.com/carbon-design-system/carbon). Had to dig a bit to find the details [here](https://carbondesignsystem.com/migrating/guide/develop/#@carbonicons-react).

It should work fine now and be good to go for a while again.

PS: Specifically annoying was that they renamed one component to `Link` that conflicts with other declarations...

### DIFF
--- a/app/auth/pages/signup.tsx
+++ b/app/auth/pages/signup.tsx
@@ -4,7 +4,7 @@ import { useFormik } from "formik"
 import { z } from "zod"
 import { useState } from "react"
 import { Switch } from "@headlessui/react"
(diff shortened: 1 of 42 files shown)


nsunami (approved):
Looks good to me! 

(Oh, `Link` would conflict with Next.js's `Link` component...)

PR #387, opened 2022-05-17
Add basic rendering of office files

chartgerink:
This PR adds basic rendering of office files using the Microsoft Office Viewer.

At the moment this adds the rendering of spreadsheets and word files - which is helpful in terms of content consumption.

Screenshot 2022-05-17 at 17 20 00
Screenshot 2022-05-17 at 17 20 36

We could also still add Powerpoint previewer, but nobody has published that yet 😊

### DIFF
--- a/app/pages/modules/[suffix].tsx
+++ b/app/pages/modules/[suffix].tsx
@@ -318,6 +318,21 @@ const Module = ({ module, mainFile, supportingRaw }) => {
             ) : (
               ""
             )}
+            {/* Preview Office files */}
+            {mainFile.mimeType === "application/vnd.ms-excel" ||
+            mainFile.mimeType ===
+              "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet" ||
+            mainFile.mimeType ===
+              "application/vnd.openxmlformats-officedocument.wordprocessingml.document" ? (
+              <iframe
+                src={`https://view.officeapps.live.com/op/embed.aspx?src=${mainFile.cdnUrl}`}
+                width="100%"
+                height="800px"
+                frameBorder="0"
+              ></iframe>
+            ) : (
+              ""
+            )}
           </div>
         ) : (
           ""


app/pages/modules/[suffix].tsx
```diff
@@ -318,6 +318,21 @@ const Module = ({ module, mainFile, supportingRaw }) => {
             ) : (
               ""
             )}
+            {/* Preview Office files */}
+            {mainFile.mimeType === "application/vnd.ms-excel" ||
+            mainFile.mimeType ===
+              "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet" ||
+            mainFile.mimeType ===
+              "application/vnd.openxmlformats-officedocument.wordprocessingml.document" ? (
+              <iframe
```
Comment (nsunami) [inline app/pages/modules/[suffix].tsx:327]: Assuming that we are supporting `.doc` file, do we need `application/msword`?

```suggestion
            mainFile.mimeType === "application/msword" || 
            mainFile.mimeType ===
              "application/vnd.openxmlformats-officedocument.wordprocessingml.document" ? (
              <iframe
```

nsunami:
I can confirm that I was able to see `.docx` and `.xlsx` files rendered after publishing! Two observations:

## 1. `.doc` file not supported?
I think `.doc` file was not included in the specs? I made a comment [here](https://github.com/libscie/ResearchEquals.com/pull/387#discussion_r878925037).

## 2. File upload fails for `.doc` and `.xls` files
I've tried uploading `.doc` and `.xls` files, but it failed on the Uploadcare:
image

nsunami (approved):
I think most people use `.docx` or `.xlsx` nowadays, and thus I assume the changes are ok for merging (not breaking). 

Do you think we should track the issues about `.doc`/`.xls` files in a separate issue?

app/pages/modules/[suffix].tsx
Comment (chartgerink) [reply app/pages/modules/[suffix].tsx:327]: Thanks @nsunami! 

`.doc` and `.xls` are not open file formats, so we don't permit the uploading of those. 🦢 When we were setting up Hypergraph we decided to go the [open file format](https://en.wikipedia.org/wiki/List_of_open_file_formats) route, and it's been implemented in the same way here for longevity purposes.

chartgerink:
> I think most people use `.docx` or `.xlsx` nowadays, and thus I assume the changes are ok for merging (not breaking).
> 
> Do you think we should track the issues about `.doc`/`.xls` files in a separate issue?

Agreed - I haven't seen a `.doc|.xls` in a long time 😊 For reference (from https://github.com/libscie/ResearchEquals.com/pull/387#discussion_r879108925):

> `.doc` and `.xls` are not open file formats, so we don't permit the uploading of those. 🦢 When we were setting up Hypergraph we decided to go the [open file format](https://en.wikipedia.org/wiki/List_of_open_file_formats) route, and it's been implemented in the same way here for longevity purposes.

PR #395, opened 2022-05-31
Dependency upgrades (patch only)

chartgerink:
This PR batches all dependency upgrades of the patch version (according to [SemVer](https://semver.org/)).

I tested this locally and there were no issues with respect to functionality.


chartgerink:
This one did not end up deploying well after all so I rolled it back. Goes to show we need the auto-deployment back up - will inspect Heroku's situation and set it up.

PR #397, opened 2022-06-04
Remove Quirrel from Heroku test deploys

chartgerink:
This PR removes Quirrel from Heroku test deploys.

This is at the same time a test to see whether the Heroku deploys of PRs are back up and properly installed - I accidentally removed the old pipeline it seems so I had to set the whole test environment up again 😅

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
 release: npx blitz prisma migrate deploy && npx blitz db seed
-web: quirrel ci && npm run start:production
+web: npm run start:production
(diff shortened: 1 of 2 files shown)


chartgerink:
AAH the UX on Heroku is such an issue! I just had a double review app, and when I tried to delete the redundant one it threw out the whole app pipeline 👎 

Retrying - again...

chartgerink:
Well that was only a few minutes of work at least - it's deploying again...!

PR #398, opened 2022-06-07
Add stale tagger action

chartgerink:
This PR adds a GitHub action to tag any and all stale issues (60 days) and pull requests (2 days). Based on the recommendation by @nsunami 💚 

It does not close anything - only tags them. This helps in triaging especially the PRs, as we have dependency upgrades coming on an ongoing basis.

Note: GitHub actions noob over here so I might've done something wrong - will run the Action in a sec to check.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,32 @@
+name: "Stale handler"
+on:
+  workflow_dispatch:
+  schedule:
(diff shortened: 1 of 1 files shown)


chartgerink:
Okay I just noted it's much easier to create the action through GitHub itself instead of manually setting it up! Scratching this PR and opening one through there.

PR #467, opened 2022-06-22
Add api route for main files

chartgerink:
This PR adds an api route for the main files of modules, making it more predictable. 

The need for this came up while trying to make a reproducible piece of output, and noticing the uploadcare URL was incomprehensible.

Specifically, the route is now as easy as:

```
/api/modules/main/[suffix]
```

✌️


chartgerink:
Merging this with admin privileges because I sort of need this for something 😅

PR #468, opened 2022-06-22
Make file extensions case insensitive

chartgerink:
This PR makes file extension limitations case insensitive.

This was a bug I noticed while trying to upload an R file 😅


nsunami (approved):
Looks good to me! I checked uploading "test.R" on my local, and it works.

By the way, the feedback for trying to upload an unsupported file may be a bit too subtle (Seeing an exclamation mark for uploading an unsupported file, I was not sure if that mark was because of the file type or some other network stuff). Just a small comment.

chartgerink:
Yes fair point! Would you mind opening up an issue so we can keep track of that? 🦦

PR #473, opened 2022-06-23
Add api for supporting files

chartgerink:
This PR adds an api route for supporting files. This is similar to #467.

This is a way to clean up the uploadcare CDN urls - which are very opaque for usage. They're also brittle in case we ever want to change the storage provider.

So with this PR you can easily retrieve files across modules using the following route that is stable even if we switch storage providers:

```
/api/modules/supporting/[suffix]/[filename]
```

An example could be (after merging this)

```
https://researchequals.com/api/supporting/kpg3-tb37/plot.png
```

This will effectively help make ResearchEquals a storage provider too - which could be helpful in the future if we want to for example host podcasts or anything 😊

### DIFF
--- a/app/api/modules/supporting/[suffix]/[filename].ts
+++ b/app/api/modules/supporting/[suffix]/[filename].ts
@@ -1,0 +1,46 @@
+import { BlitzApiHandler } from "blitz"
+import { PrismaClient, Prisma } from "@prisma/client"
+import db from "db"
+import https from "https"
+
+const handler: BlitzApiHandler = async (req, res) => {
+  const {
+    query: { suffix, filename },
+  } = req
+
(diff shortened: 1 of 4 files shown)


chartgerink:
Nice, thanks @nsunami! 🙏  I think you tested the error by trying to upload a duplicate file? 😊

nsunami:
> Nice, thanks @nsunami! 🙏 I think you tested the error by trying to upload a duplicate file? 😊

Of course!


I tested the error on this portion:

```
      if (file.original_filename == newFile.original_filename) {
        throw new Error("Please check for duplicates.")
```


But not this one:

```
          error: (error) => {
            return error.toString()
          },
```

PR #485, opened 2022-06-29
Add state management to the workspace settings

### LINKED ISSUE #235: If I start to change my information in settings and switch the settings tab, my form information is reset...
I was changing some information in my bio, and when I swapped from the "Workspace" tab of the settings modal to the "Account" tab and swapped back, the changes were reset and I lost my previously written (yet unsaved) bio information.

Perhaps there should be a prompt to save your work should you have changed anything in the form and attempt to close/change tabs.

chartgerink:
This PR adds an improved state management mechanism for the workspace settings. Fixes #235.

Previously, the workspace settings would set the current values for the bio, pronouns, etc. based on what is in the database **upon every render**. This meant that local changes did not persist when switched out of the specific window or upon refresh.

Now, with the `recoil` library we improve the state management and go through the following steps instead:
1. Initialize empty state for each of the fields (e.g., `firstName = ""`)
2. When workspace settings component is loaded, set the state with the latest information from the database (i.e., when the state is empty)
3. Write the fetched state to `localStorage`
4. Update `localStorage` with any changes made (e.g., when name is updated)
5. State is persisted until cancel button is clicked, in which case the state is reverted to empty (i.e., as in 1)

This specific implementation creates a granular state, one for each field. This could be improved in the future by collecting them in one state for workspace overall. 

This is also not the only place where state management improvements can benefit the application, but I only implement it here for now - will open up issues to track the work in for example drafts.

## Todo upon approve

- [x] Add `recoil` to dependencies in the wiki
- [x] Add `recoil-persist` to dependencies in the wiki


nsunami (approved):
Looks good to me implementation-wise! (A caution is that I don't have experience with Recoil)

One thought here is that when I navigate to another page, the values will be lost. Is that intended?

@chartgerink  I'm also curious to know how you chose this solution rather than a warning message (e.g., "Your changes will be lost"). 👀

chartgerink:
@nsunami - thanks for testing 😊 I also didn't have experience with recoil before this, so my implementation probably leaves room for improvement 😅 

For me, the state is persisted across page loads - did you test in private mode? This all depends on the `localStorage` being active.

> I'm also curious to know how you chose this solution rather than a warning message (e.g., "Your changes will be lost"). 👀

Honestly, I hate creating modals at the moment 🤷‍♀️ They feel like a lot of work both on the dev end (but aren't really) and on the user's end? I am not really anti creating it but my judgment led me to believe persisting changes made would prevent losing changes if somebody picks up and gets distracted by something else on the website.

PR #492, opened 2022-07-06
Introducing "Pay what you want"

chartgerink:
This PR implements a "Pay what you want" model for the Open Access licenses.

This provides additional ways for people to support the project while also being in line with our core value of distributing power in research. It does not raise financial barriers to publishing Open Access, but does create a slight nudge for people who have financial means to help the project survive/thrive.

It will be interesting to monitor the amounts over time.

Please note while testing that [the Stripe payment](https://github.com/libscie/ResearchEquals.com/wiki/Dependencies#stripe) cannot be completed through the automated deployment. If you'd like to test that it needs to be run locally.

## Video demo

https://user-images.githubusercontent.com/2946344/177545090-266ab891-c994-44c1-8706-5b7a6df4c63f.mov

## Additional information

This model is made possible by Stripe's ad-hoc pricing. This is where the customer sets the price, instead of the seller. 

New products are added to Stripe, one for each license, and we feed a price amount based on the user's input. This then creates a checkout session as before, instead of a `price_id` determined a priori!

See also the following pages on the Stripe docs:

- https://stripe.com/docs/products-prices/manage-prices#ad-hoc-prices
- https://stripe.com/docs/tax/products-prices-tax-categories-tax-behavior#tax-behavior

## Potential future work

- Introduce badges for people to display they've supported the project (which would require a badge system to begin with 😅 )


app/core/modals/PublishModuleModal.tsx
```diff
@@ -90,33 +91,171 @@ export default function PublishModule({ module, user, workspace }) {
...
+                            </p>
+                            <input
+                              type="text"
+                              id="name"
+                              name="name"
+                              required
+                              value={payWhat}
+                              className="placeholder-font-normal block w-full appearance-none rounded-md border border-gray-300 px-3 py-2 text-sm font-normal placeholder-gray-400 shadow-sm focus:border-indigo-500 focus:outline-none focus:ring-indigo-500  dark:border-gray-500 dark:bg-gray-800"
+                              onChange={(value) => {
+                                setPayWhat(parseInt(value.target.value) || 0)
```
tefkah, line 129: I think (if your goal is to get more people to pay you (which I think you should!)) it might be a good idea to set the default at a non-zero value, like idk 5/10 euros or something. This does sorta induce some kind of implicit expectation that you should pay, kind of like how bandcamp (used to?) do that, but I don't think that'll deter people or anything.

chartgerink: Thanks @tefkah for taking the time and making this suggestion!

I think that's a great idea. Will reach out to our testers and consider how we can test this empirically moving forward!

chartgerink: Came out with a 7-5 preference for non-zero default - zero default, respectively (based on tester feedback).

Will update the pull request to default at 5EUR 😊

chartgerink:
This PR is updated to include:

1. A default 5EUR price (can be adjusted to zero)
2. The draft window now includes the "pay what you want" price (still mentioned "Free" before)

I reached out to our lawyer as I realized this requires an update to the terms as well. That procedure will take a moment longer as we need to go through the following steps:

- [ ] Update the terms
- [ ] Send a notice to all users of the update
- [ ] Wait the legally required timeframe (I think two weeks)
- [ ] Then we can merge this PR 😅

app/core/modals/PublishModuleModal.tsx
tefkah: Awesome!

chartgerink:
This should be unblocked by end of month - before merging all the webhook things should be triple checked as they will become more complex as we integrate the collections feature.

chartgerink:
Progress is being made over in #705 🎉

chartgerink:
This is now unblocked and I made it a priority as it's old but contains important information. I'll take care of getting this back up to speed as I also created the PR originally.

chartgerink:
Okay tried rebasing this but with the amount of changes happening over time ended up going with a merge. The rebase-merge debate is real 😂 

Need to take a step back for a moment and revisit later - it seems to work right now in my local environment.

chartgerink:
Feeling pretty good about this after a bit of distance and testing the deployment on mobile. 

Let it rip!

PR #503, opened 2022-07-12
Improve state management of onboarding information

### LINKED ISSUE #264: Workspace changes not updating until refreshing the page
For the "onboarding quests", the "Add your info" one does not update if I fill out my profile, until after I refresh the page.

It would be expected that this would update automatically. This is a remaining bit of #229 and was not fixed in #257.

chartgerink:
This PR improves the state management of the onboarding information. Fixes #264.

https://user-images.githubusercontent.com/2946344/178457312-ca30ba21-ce4a-4f19-af55-626cd8c3670a.mov

This took a few more steps than I'd thought, because I needed to also manage the state of the settings modal now (otherwise the modal would close automatically once all the information was entered but not yet saved).

I am learning about the `recoil` library and this is really a fantastic way to increase the responsive feel of the platform while at the same time reducing the database actions needed. This is possible because everything is written into the `localStorage` for persistence.

### DIFF
--- a/app/core/components/OnboardingQuests.jsx
+++ b/app/core/components/OnboardingQuests.jsx
@@ -1,14 +1,23 @@
 import { useMutation, useQuery } from "blitz"
 import { Widget } from "@uploadcare/react-widget"
-import { useRef } from "react"
+import { useEffect, useRef, useState } from "react"
 import { Formik, Form } from "formik"
 import { Email, UserAvatar, User, Parameter } from "@carbon/icons-react"
+import { useRecoilValue, useRecoilState } from "recoil"
 
 import SettingsModal from "../modals/settings"
 import changeAvatar from "../../workspaces/mutations/changeAvatar"
 import getSignature from "../../auth/queries/getSignature"
 import QuickDraft from "../../modules/components/QuickDraft"
 import resendVerification from "../../auth/mutations/resendVerification"
+import {
+  workspaceFirstNameAtom,
+  workspaceLastNameAtom,
+  workspaceBioAtom,
+  workspacePronounsAtom,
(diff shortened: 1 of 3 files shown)


app/core/components/OnboardingQuests.jsx
```diff
@@ -1,14 +1,23 @@
 import { useMutation, useQuery } from "blitz"
 import { Widget } from "@uploadcare/react-widget"
-import { useRef } from "react"
+import { useEffect, useRef, useState } from "react"
```
nsunami, line 3: Do we need to load these dependencies? If not, maybe roll back to like this?
```suggestion
import { useRef } from "react"
```

app/core/components/OnboardingQuests.jsx
```diff
@@ -1,14 +1,23 @@
...
 import { Formik, Form } from "formik"
 import { Email, UserAvatar, User, Parameter } from "@carbon/icons-react"
+import { useRecoilValue, useRecoilState } from "recoil"
 
 import SettingsModal from "../modals/settings"
```
nsunami, line 8: No longer used here?

```suggestion
```

app/core/modals/settings.tsx
```diff
@@ -1,6 +1,8 @@
 import { Dialog, Transition, Tab } from "@headlessui/react"
 import { Fragment, useState } from "react"
 import { Close } from "@carbon/icons-react"
+import { useRecoilState, useResetRecoilState } from "recoil"
```
nsunami, line 4: Just removing unused dependency

```suggestion
import { useRecoilState } from "recoil"
```

PR #512, opened 2022-07-15
Upgrade Autocomplete to use createRoot

### LINKED ISSUE #326: Auto-unfocusing/blurring from the reference text input
When I click the reference input box, the drop-down suggestion appears. But, after that, the drop-down closes, and I lose focus on the input box (then, I have to click the box again):

https://user-images.githubusercontent.com/17035406/153646521-464b64be-4f68-481e-8387-bd5054bf57df.mov

Edited: 
I was using Chrome @chartgerink

### LINKED ISSUE #364: `React.render`warning in `Autocomplete.tsx`
As mentioned [here](https://github.com/libscie/ResearchEquals.com/issues/326#issuecomment-1039130069), the [`Autocomplete`](https://github.com/libscie/ResearchEquals.com/blob/main/app/core/components/Autocomplete.tsx) component uses the render method, which has been updated in React@18. 

The warning does not break anything at this point, but does prevent utilizing the full capacity of React@18. It would be helpful to resolve this warning before it becomes a breaking issue.

```
Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. 
Until you switch to the new API, your app will behave as if it's running React 17. 
Learn more: https://reactjs.org/link/switch-to-createroot
```

[Link to React docs on this](https://reactjs.org/link/switch-to-createroot). This seems to be relatively minor but my attempts didn't succeed 🥎  If you know how to fix this, we welcome a PR!

chartgerink:
This PR upgrades the `Autocomplete` component to use `createRoot` as is necessary for utilizing React 18. Fixes #364. It ended up being [documented by Algolia](https://www.algolia.com/doc/ui-libraries/autocomplete/integrations/using-react/?client=jsx#with-react-18) too.

Please note that I changed the file from `.tsx` to `.jsx` - there were some Typescript errors I did not know how to work around at this point:

```
Property 'unmount' does not exist on type 'never'.
```

So if anyone knows their Typescript and can help work around this - I would love your help!

Note: Does not fix #326 - makes it seem a tiny bit less of an issue though?

### DIFF
--- a/app/core/components/Autocomplete.jsx
+++ b/app/core/components/Autocomplete.jsx
@@ -1,9 +1,11 @@
 import { autocomplete } from "@algolia/autocomplete-js"
 import React, { createElement, Fragment, useEffect, useRef } from "react"
-import { render } from "react-dom"
+import { createRoot } from "react-dom/client"
 
 function Autocomplete(props) {
   const containerRef = useRef(null)
+  const panelRootRef = useRef(null)
+  const rootRef = useRef(null)
 
   useEffect(() => {
     if (!containerRef.current) {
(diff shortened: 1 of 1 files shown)


chartgerink:
I think this is quite the usability issue so I'd rather not merge to be honest and keep the warning.

We could fix this but I'm also not sure where to start. Seems like all of this (plus the blurring issue) is starting to make me wonder whether the Algolia implementation is the best route here.

chartgerink:
Closing this as a wontfix for now. If we ever need to come back to it we can...

PR #527, opened 2022-07-21
Add `/api/avatars/[handle]` route

chartgerink:
This PR adds an API route for avatars - this is similar to #473 and #467.

The main reason to do this is to allow for permalinking to avatars - for example like github does. This means that whenever somebody embeds an avatar based on this route, the link won't break when a person updates their avatar. This is yet another step towards a CDN pretty much 😊 

The only quirks I detected are:
- We used faker to generate fake avatar links - these don't really work because each load of the URL generates something else and it doesn't have a clear mimetype 👇 
Screenshot 2022-07-21 at 09 26 13
- I ended up not using this everywhere, as there's a slight delay in the image being loaded. This is because every link goes through our server, which isn't optimized for high amounts of request (e.g., 30 avatars at once on `/browse`). I only implemented it on `/[handle]` as a result

I'm not 100% sure yet I want it in, but I had the thought of implementing this during the week and wanted to at least document the work!

@nsunami would much appreciate your opinion/questions/feedback on this - what do you think?

### DIFF
--- a/app/api/avatars/[handle].ts
+++ b/app/api/avatars/[handle].ts
@@ -1,0 +1,39 @@
+import { BlitzApiHandler } from "blitz"
+import db from "db"
+import https from "https"
+
+const handler: BlitzApiHandler = async (req, res) => {
+  const {
+    query: { handle },
+  } = req
+
+  const author = await db.workspace.findFirst({
+    where: { handle: handle?.toString() },
+  })
+  return new Promise((resolve, reject) => {
+    https
+      .get(author?.avatar!, (response) => {
+        var data = [] as any
+
+        let buffer
+        response
+          .on("data", function (chunk) {
+            data.push(chunk)
+          })
+          .on("end", function () {
+            res.statusCode = 200
+            res.setHeader("Content-Type", "image")
+            res.setHeader("Content-Disposition", `filename=avatar-${author?.handle}`)
+            var buffer = Buffer.concat(data)
+            res.end(buffer)
+            resolve()
+          })
+      })
+      .on("error", (e) => {
+        res.status(404).end()
+        resolve()
+      })
+  })
+}
+
+export default handler
--- a/app/api/avatars/[handle].ts
+++ b/app/api/avatars/[handle].ts
@@ -10,6 +10,8 @@ const handler: BlitzApiHandler = async (req, res) => {
   const author = await db.workspace.findFirst({
     where: { handle: handle?.toString() },
   })
+
+  const mimetype = !author?.avatar!.match(/ucarecdn/g) ? "image/svg+xml" : "image"
   return new Promise((resolve, reject) => {
     https
       .get(author?.avatar!, (response) => {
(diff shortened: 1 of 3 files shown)


nsunami:
I like the direction it's going! 

I see that when a user does not have an avatar set, the page still looks for an image (and thus breaks the avatar, as shown in the screenshot). Can we catch this, and continue to show the [UI avatars](https://eu.ui-avatars.com/) for users without uploaded avatars?

chartgerink:
Ah wait this is a quirk in the testing environment I think. 

The avatar is set but the URL from faker isn't a regular image.

All users in production have an image URL - either their own or with the initials. I updated the seed script in this PR as well so it doesn't happen in test either :)

nsunami:
What can we do to fix this? 
I played around a bit and found that accessing the API endpoint without an image produces an SVG.
For example, accessing `/api/avatars/adele.parisian` would give me:
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="64px" height="64px" viewBox="0 0 64 64" version="1.1">
  <circle fill="#574cfa" cx="32" width="64" height="64" cy="32" r="32"/>
  <text x="50%" y="50%" style="color:#ffffff;line-height:1;font-family:-apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif" alignment-baseline="middle" text-anchor="middle" font-size="28" font-weight="400" dy=".1em" dominant-baseline="middle" fill="#ffffff">BE</text>
</svg>
```

image

But then it's not rendering..
image


Looking at source, I cannot figure out what's wrong:

```
<img src="/api/avatars/adele.parisian" class="max-w-28 h-28 max-h-28 w-28 rounded-full border border-2 border-gray-900 dark:border-white " alt="Avatar of adele.parisian">
```

chartgerink:
Thanks for your patience on this one @nsunami  🙏 

I checked it out again and it looked like the mimetype had to be dynamic. That is `image/svg+xml` when it returned the svg as you showed. Otherwise it just uses `image` and that is fine for most but not svg.

Once you review this change I think we can merge 😊

nsunami:
@chartgerink I checked the review app, and I'm seeing that the avatars are still appearing as I saw last time:
image


This time, accessing the avatars API endpoint (`/api/avatars/garnett.bode`) gives me an empty XML file (no trees).
 
image

chartgerink:
Thanks @nsunami!

The seeded database has links to things that aren't images - they're odd in terms of mimetype. I mentioned this in a previous edit of a comment up top (although stricken through so did not expect you to read it!): 

>All users in production have an image URL - either their own or with the initials. I updated the seed script in this PR as well so it doesn't happen in test either moving forward :)

So this is not unexpected right now. See also the changes to the [`db/seeds.ts`](https://github.com/libscie/ResearchEquals.com/pull/527/files#diff-67a0a6bf95090b17083d15089c6398041789603bfef9ac8d56289d5ec9946131) file in the PR.

Would you be okay with us moving forward on this based on this explanation?

nsunami (approved):
Got it! Then, LGTM!

PR #529, opened 2022-07-21
Fix encoding of filenames in Content-Disposition

### LINKED ISSUE #528: 502 bad gateway after trying to open supporting file
In a recently published module something is happening that creates an error on the server side. 

Digging in now and will update... The server does recover from this error after a few minutes but this needs to be fixed 

Sorry @GNilsonne for this unexpected issue!

chartgerink:
This PR adds encoding of the filename in the `Content-Disposition` of the following API routes:

- `/api/modules/supporting/[suffix]/[filename]`
- `/api/modules/main/[suffix]`

Fixes #528.

### DIFF
--- a/app/api/modules/main/[suffix].ts
+++ b/app/api/modules/main/[suffix].ts
@@ -22,7 +22,7 @@ const handler: BlitzApiHandler = async (req, res) => {
           .on("end", function () {
             res.statusCode = 200
             res.setHeader("Content-Type", module?.main!["mimeType"])
-            res.setHeader("Content-Disposition", `filename=${module?.main!["name"]}`)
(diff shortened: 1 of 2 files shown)


nsunami:
@chartgerink I feel like this is a good case for a regression testing—do you think it's feasible, and it's worth to write one? (I don't have much testing experiences but happy to explore one)

chartgerink:
> @chartgerink I feel like this is a good case for a regression testing—do you think it's feasible, and it's worth to write one? (I don't have much testing experiences but happy to explore one)

It would be great to add - my test suite knowledge is rather limited. I think Blitz defaults to Jest, but I haven't tried much of that (there's just the default tests now). 

Continue on Discord? :-)

Invite if anybody else reads this and wants to join too

https://discord.gg/SefsGJWWSw

PR #530, opened 2022-07-22
Improve mimetype recognition + add powerpoint previews

chartgerink:
This PR improves mimetype recognition in the preview logic.

I discovered this bug based on a recent module that was shared on the platform, but was not displaying a preview (despite being a PDF file). 

I looked into it and it appeared the mimetype was `application/pdf; charset=utf-8` whereas the logic operated on `application/pdf`. This fix doesn't equate the mimetype to `application/pdf` but checks whether the mimetype starts with it. This fix is also applied to other mimetypes to prevent this issue.

While I was at it I also added the mimetype for powerpoint previews.

### DIFF
--- a/app/pages/modules/[suffix].tsx
+++ b/app/pages/modules/[suffix].tsx
@@ -283,7 +283,7 @@ const Module = ({ module, mainFile, supportingRaw }) => {
               ""
             )}
             {/* Preview PDF */}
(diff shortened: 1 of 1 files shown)


chartgerink:
Mergin this as this has been open for a few days and is priority - the PDF viewer didn't function 100% and I'd like to restore it with this fix.

PR #538, opened 2022-07-25
Remove workspaces without memberships and authorships

### LINKED ISSUE #475: Remove workspaces without members
Currently, we remove accounts that have not verified their email and have been inactive after 30 days. That is set up over in [/app/api/remove-unverified-accounts.ts`](https://github.com/libscie/ResearchEquals.com/blob/main/app/api/remove-unverified-accounts.ts).

At the moment, this does not remove the workspaces they were a member of. This issue tracks the implementation of deleting the workspaces, but only those that fullfill all the following conditions:

1. Do not have any members
2. Do not have any published modules

This way we ensure the database is cleaned up regularly and we do not get a sense of overinflated workspace numbers either.

## Workspaces v accounts

Wait - what is the difference between a workspace and an account you ask? Good question!

- An account (or user) is with which you log in. 
- A workspace is the authoring entity

Right now, when you sign up your account gets linked to your personal workspace. We may implement the option for people to be a part of multiple workspaces in the future, with other people even. Let me know if this is of interest to you!

chartgerink:
This PR adds an additional call to the CRON job that runs every 1st day of the month, to also delete workspaces that have no memberships (i.e., no linked accounts) and no authorships (no drafts or published modules).

This cleans out the authorspace, primarily in the following scenario:
1. Somebody signs up
2. Doesn't verify email in 30 days
3. Account gets deleted
4. Workspace now would remain, with this PR it will also be deleted if there are no drafts.

Fixes #475.

### DIFF
--- a/app/api/remove-unverified-accounts.ts
+++ b/app/api/remove-unverified-accounts.ts
@@ -50,6 +50,17 @@ export default CronJob(
       })
     })
 
-    //  TODO: Remove workspaces without memberships and authorships
+    // Remove workspaces without memberships and authorships
+    // This helps clean up any unused author profiles.
+    await db.workspace.findMany({
+      where: {
+        members: {
+          none: {},
+        },
+        authorships: {
+          none: {},
+        },
+      },
+    })
   }
 )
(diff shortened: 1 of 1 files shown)


nsunami (approved):
LGTM! 

Just to confirm—the reason why we are not cascade-deleting the users is that we want to retain workspaces in case they have memberships or authors hips, right?

chartgerink:
Exactly! 😄

nsunami:
> Exactly! 😄 

Great!

PR #539, opened 2022-07-25
Add language to module

### LINKED ISSUE #363: Language option for modules
Currently, modules from all languages are put on to the same pile - with users publishing primarily in English and some in Dutch. One of the selling points of ResearchEquals is that authors get to choose which language they publish in.

In order to support this better, we could include the language into the module metadata. Both in terms of:

* Database schema
* CrossRef metadata deposition
* Algolia search index

The languages available in the CrossRef metadata may be the limiting factor in the options we provide here, and if I understand correctly these are the [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) language codes (there are [modules to easily deal with these](https://www.npmjs.com/package/iso-639-1)).

## Requirements

- [x] Update database schema
- [x] Add option to select language to quick draft
- [x] Add option to edit language in draft
- [x] Add indexing of language into algolia
- [x] Update CrossRef XML generation to include language (see [schema](https://data.crossref.org/reports/help/schema_doc/5.3.1/index.html))
- [ ] Add metadata to already published modules (manual; where possible)
- [ ] Update CrossRef deposits for already published modules (manual by superadmin)

chartgerink:
This PR adds the language code to modules.

Specifically, it uses the ISO 639-1 standard as implemented in the [`iso-639-1` npm package](https://www.npmjs.com/package/iso-639-1). The parent standard is also indicated in the [CrossRef metadata schema](https://data.crossref.org/reports/help/schema_doc/5.3.1/index.html).

Screenshot 2022-07-25 at 14 11 19

This is still a work in progress, but it's ongoing in any case. 

When ready and merged: Fixes #363.

### DIFF
--- a/app/modules/components/QuickDraft.tsx
+++ b/app/modules/components/QuickDraft.tsx
@@ -6,6 +6,7 @@ import { useFormik } from "formik"
 import { Fragment, useState } from "react"
 import { z } from "zod"
 import { Checkmark, Close, InformationSquareFilled } from "@carbon/icons-react"
+import ISO6391 from "iso-639-1"
 
 import createModule from "../mutations/createModule"
 import toast from "react-hot-toast"
@@ -23,6 +24,7 @@ const QuickDraft = ({ buttonText, buttonStyle, refetchFn }) => {
       main: "",
       type: "",
       license: "",
+      language: "en",
       displayColor: "#574cfa",
     },
     validate: validateZodSchema(
@@ -31,6 +33,7 @@ const QuickDraft = ({ buttonText, buttonStyle, refetchFn }) => {
         description: z.string(),
         type: z.string().min(1),
         license: z.string().min(1),
+        language: z.string().min(1).max(2),
         displayColor: z.string().min(1),
       })
     ),
@@ -41,6 +44,7 @@ const QuickDraft = ({ buttonText, buttonStyle, refetchFn }) => {
           description: values.description,
           typeId: parseInt(values.type),
           licenseId: parseInt(values.license),
+          language: values.language,
           authors: [],
           displayColor: values.displayColor,
         }),
@@ -82,6 +86,7 @@ const QuickDraft = ({ buttonText, buttonStyle, refetchFn }) => {
     formik.setFieldValue("main", "")
     formik.setFieldValue("type", "")
     formik.setFieldValue("license", "")
+    formik.setFieldValue("language", "en")
     formik.setFieldValue("displayColor", "#574cfa")
   }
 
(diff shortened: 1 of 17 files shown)


app/modules/components/QuickDraft.tsx
```diff
@@ -31,6 +33,7 @@ const QuickDraft = ({ buttonText, buttonStyle, refetchFn }) => {
         description: z.string(),
         type: z.string().min(1),
         license: z.string().min(1),
+        language: z.string().min(1).max(2),
```
nasainsbury, line 36: Apologies if you've already considered this/is wrong, but could you do:

`language: z.enum(ISO6391.getAllCodes()),`?

nasainsbury: Actually, just tried it and am getting some quirky TS errors, will have a look.

chartgerink: I didn't know about this method - that would be much more precise validation indeed! 

Thanks for teaching me about this :)

chartgerink:
Noticed some discrepancies between the CrossRef schema and the ISO 639-1 standard, reported over on CrossRef's gitlab repository -> https://gitlab.com/crossref/schema/-/issues/19

Four languages are affected by this, which would result in failed DOI minting:

- `jv` - Javanes
- `mh` - Marshallese
- `zu` - Zulu

app/modules/components/QuickDraft.tsx
chartgerink: I've opened an issue for this to track the work (see #544). 

This PR is ready for review otherwise 😊

nasainsbury: OK sweet, nice one :)

nasainsbury:
Looks good to me, was just thinking about the language option in the DB. As far as I can tell (and I could be wrong), the DB just accepts a string, in theory a user could change the value of the input to something random, i.e `kjhasdh`, and submit, it would be saved. Is that fine for now (just wasn't sure on the acceptance criteria)? Will also take a look into displaying the language on the module page too if that's OK

chartgerink:
@all-contributors add @nasainsbury for code and design

nasainsbury (approved):
LGTM, had a look through + my commit. I'm missing context though probably :)

nsunami (approved):
LGTM!

chartgerink:
Perfect - thanks @nsunami and @nasainsbury for helping realize this feature! 

Will merge now as we have worked through the feature 🥳

PR #551, opened 2022-07-27
Add algolia updatedAt indexing

### LINKED ISSUE #505: Sort workspaces in the search bar
Currently the search bar returns authors and modules, where the modules are sorted by publication date. This is a dynamic sort and allows for new things to be shown over time.

However, the authors are sorted statically and always show the same people:

Screenshot 2022-07-12 at 13 34 08

It may be a nice idea to sort the authors on most recent activity to the profile, and permit a bit more spontaneous discovery based on that.

This would require:

- [ ] Adding the `updatedAt` property to Algolia for workspaces when indexing happens
- [ ] Adding the sort based on ☝️ in Algolia

Additionally we could add a `cron` job to simply update the index in Algolia every so often, using quirrel - this way we can update the current records easily, and not just for new workspace updates.

chartgerink:
This PR adds the `updatedAt` property to Algolia for workspaces. This happens when:

1. Email is verified and a workspace is indexed for the first time
2. Every midnight

Fixes #505.

### DIFF
--- a/app/api/update-index.ts
+++ b/app/api/update-index.ts
@@ -1,0 +1,24 @@
+import db from "db"
+import { CronJob } from "quirrel/blitz"
+import algoliasearch from "algoliasearch"
+import moment from "moment"
+
+const client = algoliasearch(process.env.ALGOLIA_APP_ID!, process.env.ALGOLIA_API_ADMIN_KEY!)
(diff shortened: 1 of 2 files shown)


chartgerink:
Merging without review as the review has become stale as far as I can tell.

PR #556, opened 2022-07-28
Remove @stripe/stripe-js

chartgerink:
Replaces  #554.

This dependency was previously marked for deletion - and after testing it locally it seemed like it was indeed unnecessary. The `stripe` dependency is sufficient.

## To do
- [x] Remove from wiki

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,6 @@
         "@headlessui/react": "1.6.6",
         "@prisma/client": "4.1.0",
         "@react-pdf-viewer/core": "3.6.0",
-        "@stripe/stripe-js": "1.32.0",
         "@tailwindcss/forms": "0.5.2",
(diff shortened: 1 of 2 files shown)


nsunami (approved):
LGTM!

@chartgerink On a side note—is there any way to test strip on a test app? (or testing on local is the only way now?)

chartgerink:
Thanks @nsunami!

As [per the wiki](https://github.com/libscie/ResearchEquals.com/wiki/Dependencies#stripe), there's no way to test this in the Heroku deployment at the moment because:

> Cannot be tested with automated deployments, because this needs Stripe webhooks to be active. Because Heroku deployment apps have a priori not known URLs, we do not have a priori accepted webhooks.

This necessitates running it locally as a result.

PR #559, opened 2022-07-29
Add FilePreviewer to drafts

### LINKED ISSUE #488: Implement a file viewer on the drafts page
I was wondering if implementing a file viewer on the drafts page would be a nice enhancement.

Currently, we cannot see a file uploaded to a draft:

image

Of course, we can always download the file and check it. But, since we have a file viewer for the published modules, it would also make sense to implement here. People may want to preview how it would look like before publishing. 

Thought about this while working on #481.

chartgerink:
This PR adds a file previewer to the drafts page.

In this change, I factored out the already existing preview code for the published page into it's own component. This is now reused in the `/drafts` page - upgrades to file previews will that way also upgrade in both places 😊 

I'm closing out July's milestones so I might end up merging this without review if deployment and testing works.

Fixes #488.

### DIFF
--- a/app/core/components/FilePreviewer.tsx
+++ b/app/core/components/FilePreviewer.tsx
@@ -1,0 +1,113 @@
+import { Viewer, Worker } from "@react-pdf-viewer/core"
+import ReactMarkdown from "react-markdown"
+import remarkGfm from "remark-gfm"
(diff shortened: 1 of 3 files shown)


chartgerink:
Merging this after testing in deployment 😊 

Great suggestion @nsunami - really helps improve the platform and allows us to further improve the drafting experience in the future.

PR #568, opened 2022-08-03
Show supporting files instead of downloading, when clicking their title 

### LINKED ISSUE #566: One-click preview for supporting files?
Related to the discussion #565, and as I raised in the issue #488, currently clicking the title of the supporting file would download the file 
image

This behavior is due to the `download` attribute for the `a` tag.

https://github.com/libscie/ResearchEquals.com/blob/eddbcbce3697a933ea6e2fabf11ea16fc365a650/app/modules/components/ViewFiles.tsx#L7-L13

I think removing it would make sense, since clicking the title will open the file in a new tab. This way, the user has to perform one click to view a file.

nsunami:
This PR will make clicking the supporting title show the file, instead of downloading. Fixes #566


chartgerink (approved):
Thanks @nsunami for opening the issue and proposing a fix here.

For any file that doesn't have a previewer in the browser, it will still download the file (e.g., `.R` file). Seems to make sense.

Feel free to merge at your convenience.

PR #580, opened 2022-08-05
Wrap the edit icon on the draft page as a button

### LINKED ISSUE #579: The edit button does not have other states and thus does not look actionable
Currently, the edit icon of the draft page does not have states for hover, active, or focus state, giving an impression that it's non-actionable.

 
Location:
https://www.researchequals.com/drafts

Icon:
image

I expect to have a pointer cursor when I hover upon the icon.

nsunami:
This PR wraps the edit icon in a `button` element so that it appears as a button. This will enable focus and hover:cursor-pointer by itself (+ has an added benefit of focusing by tab).

Fixes #579.

### DIFF
--- a/app/modules/components/ModuleEdit.tsx
+++ b/app/modules/components/ModuleEdit.tsx
@@ -333,25 +333,27 @@ const ModuleEdit = ({
           </div>
         </div>
         <div className="items-middle pt-8">
-          {isEditing ? (
-            <EditOff
-              size={24}
-              className="h-6 w-6 fill-current text-gray-900 dark:text-gray-200"
-              onClick={() => {
-                setIsEditing(false)
-              }}
-              aria-label="End editing mode without saving"
-            />
-          ) : (
-            <Edit
-              size={24}
-              className="h-6 w-6 fill-current text-gray-900 dark:text-gray-200"
-              onClick={() => {
-                setIsEditing(true)
-              }}
-              aria-label="Start editing mode"
-            />
-          )}
+          <button>
+            {isEditing ? (
+              <EditOff
+                size={24}
+                className="h-6 w-6 fill-current text-gray-900 dark:text-gray-200"
+                onClick={() => {
+                  setIsEditing(false)
+                }}
+                aria-label="End editing mode without saving"
+              />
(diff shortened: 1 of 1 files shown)


app/modules/components/ModuleEdit.tsx
```diff
@@ -333,25 +333,27 @@ const ModuleEdit = ({
...
+              <Edit
+                size={24}
+                className="h-6 w-6 fill-current text-gray-900 dark:text-gray-200"
+                onClick={() => {
+                  setIsEditing(true)
+                }}
+                aria-label="Start editing mode"
+              />
+            )}
+          </button>
```
chartgerink, line 356:
```suggestion
          <button onClick={() => {
                  setIsEditing(!isEditing)
                }} aria-label={isEditing ? "End editing mode without saving" : "Start editing mode"}>
            {isEditing ? (
              <EditOff
                size={24}
                className="h-6 w-6 fill-current text-gray-900 dark:text-gray-200"
              />
            ) : (
              <Edit
                size={24}
                className="h-6 w-6 fill-current text-gray-900 dark:text-gray-200"
              />
            )}
          </button>
```

chartgerink (changes requested):
Thanks for the proposed fix @nsunami! 👍 

Could you move the `aria-label` and `onClick` up one level as in the suggested change? I can focus on the button while testing but it doesn't do anything at the moment. The change would need to be linted because I did it only in the browser.

Happy to take it from here if you're too busy as well 😊

app/modules/components/ModuleEdit.tsx
nsunami: Applied the changes!

nsunami:
Thanks, @chartgerink! I applied your changes 👍

chartgerink:
Hm I don't see changes here @nsunami - did something maybe go wrong in pushing the changes to remote? I still see the `onClick` even on the icon level.

nsunami:
@chartgerink Sorry I forgot to push my change! It should be here now.

chartgerink (approved):
Thanks @nsunami 😊

PR #585, opened 2022-08-08
Increase stale PR to 10 days

chartgerink:
Marking PRs as stale after 2 days felt like it was a bit much. This PR proposes to change it to 10 days and see how that feels.

The auto-deployments will self-destruct after two days regardless.

@nsunami any thoughts?


nsunami:
@chartgerink I think 2 days are good given our current cadence.

nsunami (approved):
LGTM!

PR #651, opened 2022-08-29
Add collections announcer

chartgerink:
This PR adds some announcements/teasers for the upcoming collections feature, due for September development. 

It's intentionally vague about launch date, although it is planned to be launched on October 1st. This is because hard promises may not be fulfilled in case something unexpected happens. 

This also introduces the pricing model, which went through some last minute changes. 

## Screenshots

Screenshot 2022-08-29 at 14 51 57
Screenshot 2022-08-29 at 14 51 58

## Note

I double checked whether we can issue DOIs for collections as journal issues - [the example XML](https://gitlab.com/crossref/schema/-/blob/master/best-practice-examples/journal.vol.issue5.3.0.xml) verifies it is possible.


nsunami:
Looking great! Two thoughts:

1. Clicking around the options cards changes the positions.

https://user-images.githubusercontent.com/17035406/187463445-992e6c54-9557-4e4d-9d87-6c82b5b7ecd7.mov


2. I did not realize at first that the close button is there at the right upper hand corner. Would it be a good idea to make the "Coming soon..." button clickable and do closing there?

nsunami (approved):
LGTM! I made small comments above. I will leave it up to you if you want to implement them or not.

chartgerink:
Thanks @nsunami! I will create two new issues to track fixes for these tomorrow 😊

Issue #677, opened 2022-09-08
Feature: Collections 


chartgerink:
In an attempt to track the work for the Collections feature, I'm creating this overarching issue. 

Sorry in advance that this is likely to not be super accessible for collaboration - I am on a tight schedule (two weeks) to get this feature implemented and tested, which means sometimes the documentation in the issues will be sparse/lacking. I am a team of one for this feature right now, so that's what it is 🤷‍♀️ 

Regardless, I don't want to fully exclude the potential for some discussion. Please do feel free to raise questions. If you see something going on and would like to discuss, hop into [the audio lounge on Discord](https://discord.gg/SefsGJWWSw) - I am more likely to be available for a quick chat there.

### Workflows

- [x] Create a collection
  - [x] #681
  - [x] Collaborative collection creation
  - [x] Community collection creation
  - [x] #691
- [x] Upgrade a collection
  - [x] Free -> Collaborative
  - [x] Collaborative -> Community
  - [x] Free -> Community
- [x] #712
- [x] Administer collection
  - [x] Add editor(s)
  - [x] Add item to collection
    - [x] Add comment
  - [x] Go through submitted items (Community only)
  - [x] Change default sort
  - [x] Change pinned items
  - [x] Customize collection
    - [x] Title (Collaborative, Community)
    - [x] Icon (Collaborative, Community)
    - [x] Header image (Community)
- [x] #718
- [x] Submit to Community collection
- [ ] #717

### Routes

- [x] #710
- [x] #711
- [x] #700

### Back-end

- [x] #680
  - [x] Collections
  - [x] Submissions
  - [x] Editors
  - [x] Contributors
- [x] #731
  - [x] #684
  - [x] Resolve PR #614 
  - [x] #699
- [x] #683
  - [x] Test products
  - [x] Add test products to seeding script
  - [x] Real products
  - [x] Add real products to seeding script
  - [x] Add upgrades of collections
    - [x] Test environment
    - [x] Production environment
- [x] #690
- [x] Stripe webhook upgrades
  - [x] #692
  - [x] Process upgrades of collection in `stripe_webhook`

Issue #546, opened 2022-07-26
Remove redundant ternary operators


chartgerink:
A ternary operator is ([according to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator))

> ... the only JavaScript operator that takes three operands: a condition followed by a question mark ( ? ), then an expression to execute if the condition is truthy followed by a colon ( : ), and finally the expression to execute if the condition is falsy.

There's a lot of ternary operators spread throughout the application that follow this pattern:

```
x === "something" ?
   <>content</> :
   ""
```

Based on feedback from @nasainsbury, it seems like we could simplify the code in those cases:

```
x === "something && <>content</>
```

This would reduce the amount of (potential) confusion in the ternary logic. Unchanged functionality would need to be checked upon implementation. 

This is a great issue to work on if you'd like to contribute to the project but don't want to dive in real deep. This contribution will still benefit the project and we'd much appreciate if you choose to do it 🙏

nsunami:
During the review on the collections feature, I found a similar place for optimization (mentioned https://github.com/libscie/ResearchEquals.com/pull/685#discussion_r977794577)

chartgerink:
Closing because this has gotten stale.

PR #685, opened 2022-09-11
Add collections feature

chartgerink:
A second try at all the work documented in #677 😊 

Ongoing work - some moral encouragement for future Chris who revisits this page:


app/collections/queries/getCollection.ts
```diff
@@ -0,0 +1,117 @@
...
+        where: {
+          id: ctx.session.$publicData.workspaceId,
+        },
+      },
+    },
+  })
+
+  let isAdmin = false
+  if (ctx.session.$publicData.workspaceId) {
+    console.log("yes")
```
nsunami, line 91: I'm just double-checking that this logging is intended! If so, would it be useful to make it a more informative? 
(e.g., `console.log("Workspace ID is present")` etc)?

app/pages/collections/[suffix]/admin.tsx
```diff
@@ -0,0 +1,417 @@
...
+              },
+              templates: {
+                item({ item, components }) {
+                  return (
+                    <>
+                      {item.__autocomplete_indexName.match(/_modules/g) ? (
+                        <SearchResultModule item={item} />
+                      ) : (
+                        ""
+                      )}
```
Comment (nsunami) [inline app/pages/collections/[suffix]/admin.tsx:335]: Would using the short-circuit operator, instead of the ternary, work the same way?

```suggestion
                      {item.__autocomplete_indexName.match(/_modules/g)  && 
                        <SearchResultModule item={item} />
```

app/collections/components/AdminEditors.tsx
```diff
@@ -0,0 +1,150 @@
...
+                  }
+                )
+              }}
+              defaultValue={editor.role}
+              className="placeholder-font-normal block appearance-none rounded-md border border-gray-400 bg-white px-4 py-2 pr-6 text-sm font-normal placeholder-gray-400 shadow-sm focus:outline-none focus:ring-2 focus:ring-indigo-500 focus:ring-offset-0 dark:border-gray-600 dark:bg-transparent dark:text-gray-200 "
+            >
+              {Object.values(MembershipRole).map((role) => {
+                return <option key={role}>{role}</option>
+              })}
+            </select>
```
nsunami, line 57: I noticed that I can still select a role that's not allowed (for example, being a user for my own collection). 

image

We don't have to fix this for this PR, but something we want to address later? I think we can make a stateful value and make this dropdown controlled to manage the value (so that the selection goes back to "OWNER" even the user selects "USER")

nsunami (approved):
LGTM! ✅ 

I had small comments here and there, but they are not about main functionality, and thus I'm good with these merged to the main! 🥳 Leave it up to you @chartgerink

app/collections/queries/getCollection.ts
chartgerink: lol no [chesterton's fence](https://fs.blog/chestertons-fence/) here! removed :)

app/pages/collections/[suffix]/admin.tsx
Comment (chartgerink) [reply app/pages/collections/[suffix]/admin.tsx:335]: We could I think - we need to do a cleanup of this throughout the app and make sure to test. This is noted in #550 - @nasainsbury mentioned willing to work on this - maybe reach out to them on Discord?

app/collections/components/AdminEditors.tsx
chartgerink: Yes! I was doing some hacking of this but there's 100% improvements - could you open that issue to track it? 👍

app/collections/queries/getCollection.ts
nsunami: Didn't know about this—cool

app/pages/collections/[suffix]/admin.tsx
Comment (nsunami) [reply app/pages/collections/[suffix]/admin.tsx:335]: Sounds good. We have an issue to track this, so I'm going to add it there too: https://github.com/libscie/ResearchEquals.com/issues/546

app/collections/components/AdminEditors.tsx
nsunami: Sounds good—here's the opened issue! https://github.com/libscie/ResearchEquals.com/issues/742

PR #720, opened 2022-09-19
Hotfix CR API parameter

### LINKED ISSUE #719: CrossRef calls failing 🐛
### URL of the Page

_No response_

### What happened?

All API calls to CrossRef are failing. There's an issue in the API call, already found the fix. Will PR soon.

### Steps to reproduce

_No response_

_No response_

### Anything else?

_No response_

chartgerink:
Fixes #719 - made a mistake I missed in the query parameter....


nsunami (approved):
Checked that now we can add a reference by DOI:

image

LGTM! ✅

PR #726, opened 2022-09-20
Delete the duplicate supporting file query and use `invoke` instead

### LINKED ISSUE #678: Duplicate `getSupportingFiles.ts` 🐛
### URL of the Page

(build phase)

### What happened?

When building the app (`blitz build`), the process warns that there are duplicate pages:

```
warn  - Duplicate page detected. /app/modules/mutations/getSupportingFiles.ts and 
/app/modules/queries/getSupportingFiles.ts both resolve to /api/rpc/getSupportingFiles.
```

It looks like there are two `getSupportingFiles.ts`: one in the [/queries ](https://github.com/libscie/ResearchEquals.com/blob/main/app/modules/queries/getSupportingFiles.ts) and the other in the [/mutations](https://github.com/libscie/ResearchEquals.com/blob/main/app/modules/mutations/getSupportingFiles.ts). I checked the diff of these files, and they seem like they are identical.

I'd think it would be a good idea to consolidate it into one. Since the function of the file is a query, it may make sense to retain the one in the query folder, and delete the other one. Then, we need to check the imports for other files.

### Steps to reproduce

Run `blitz build`

_No response_

### Anything else?

_No response_

nsunami:
This PR deletes the duplicate supporting file query (`getSupportingFiles.ts`) in the `/mutations` folder. The file being used now is in the `/queries` folder, since the function performs a query.

I also updated how we get the information about a new supporting file, by using `invoke` instead of `useMutation` hook. 

Fixes #678


chartgerink (approved):
Looks good to me - I would appreciate a bit more information like when you change from `useMutation` to `invoke` as I had to go read the docs about why this would work similarly.

Please note that this use of a query is now inconsistent with the rest of the app. It would help if we made it consistent here (with `useQuery`) or start using `invoke` elsewhere too.

nsunami:
@chartgerink Thanks for the comment Chris!

In retrospect, I think this PR scope-creeped on me, since it changes two things at the same time. It would have been a lot simpler if I had separated this into two.

Issue #724, opened 2022-09-20
Add `db seed` to Heroku test deploys


chartgerink:
We recently had a hick-up in merging PRs, where something resulted in an error in production. 

One reason was that the production environment failed during `blitz db seed`, which is not run in the Heroku deployments right now. This means that the auto-deploy made us overconfident.

This can be resolved by adding `npx blitz db seed` to the `./Procfile`.

nsunami:
Taking a note about why the app deploys on Heroku but not on AWS/`flightcontrol`. 📓

Currently, the production behavior is determined by the `flightcontrol.json` file. This script applies the db migration, seeding and building (`npx blitz prisma migrate deploy && npx blitz db seed && npm run build`). On the other hand, the Heroku script (`./Procfile`) used to only migrates the db, but do not do seeding  `npx blitz prisma migrate deploy`. Synching up these script would minimize the discrepancy between the Heroku review app environment and the production environment. 


On a side note, I wonder if it would be useful to do the db migration and seeding when testing on local. My initial idea is to add  `npx blitz prisma migrate dev && npx blitz db seed` (not deploy) into the `dev` command in the `package.json` file. But, I'm not 100% sure if that's possible & recommended.

PR #733, opened 2022-09-21
Update the `faker` seeding script

### LINKED ISSUE #621: Faker deprecation update
The `faker` library is deprecating one of the functions in the next major release according to logs:

```
[@faker-js/faker]: faker.name.findName() is deprecated since v7.4 and will be removed in v8.0. 
Please use faker.name.fullName() instead..
```

Fixing this before then would be an easy but good idea! 🙌

nsunami:
This PR updates the `faker` seeding script for the test data, since the `faker` package is deprecating the function we are using (`faker.name.findName()`). I updated the script a bit so that we will get fake first names for the `firstName` field, and fake last name for the `lastName` field.

Fixes #621 

image

-----
On a side note, I noticed that the avatar image does not correspond to the handle. I realized that this is an expected behavior given that the seeding script is generating `avatar` independent of the handle:

```js
avatar: `https://eu.ui-avatars.com/api/?rounded=true&background=574cfa&color=ffffff&name=${faker.internet.userName()}`,
```

### DIFF
--- a/db/seeds.ts
+++ b/db/seeds.ts
@@ -390,8 +390,8 @@ const seed = async () => {
                   create: {
                     handle: faker.internet.userName().toLowerCase(),
                     avatar: `https://eu.ui-avatars.com/api/?rounded=true&background=574cfa&color=ffffff&name=${faker.internet.userName()}`,
-                    firstName: faker.name.findName(),
-                    lastName: faker.name.findName(),
+                    firstName: faker.name.firstName(),
+                    lastName: faker.name.lastName(),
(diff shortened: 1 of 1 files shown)


nsunami:
Note that it may make sense to test this after we add the seeding script for Heroku (After #724)

nsunami:
Now that #724 is merged, I'm marking this ready for review ✌️

chartgerink (approved):
Straightforward and important fix, thanks 🙏

nsunami:
Natuurlijk! 👍

PR #734, opened 2022-09-21
Fix the height of the draft container to 100% to prevent the white space appearing at the bottom

### LINKED ISSUE #696: A blank space below the draft page when uploading a tall main file   🐛
### URL of the Page

https://www.researchequals.com/drafts

### What happened?

When uploading a tall main file (such as a tall image) in the draft view, there will be a blank space below the screen. 

We may want to check the overflow. 

### Steps to reproduce

1. Add a draft
2. Upload a tall image (The test image I used was 1242 × 2208)
3. Scroll down the screen (I had to scroll on the sidebar. Not on the image itself).

Chrome

### Anything else?

I noticed this while working on the Dependabot upgrade for Uploadcare (https://github.com/libscie/ResearchEquals.com/pull/653)

researchequa-dependabot-1fdhjo herokuapp com_drafts_suffix=mpk0-yp8w(iPad Mini) (2)

nsunami:
This PR fixes the height of the draft container to 100%, to fix the issue of a white space appearing below the page when the main file is tall. 

@chartgerink I was not 100% sure of the reasoning behind the existing code to calculate the height of the element (`calc(100vh - 73px - 55px)`) when the window is big. Removing that and just using `h-full` seems to be working. That said, it's possible that I may be missing something—let me know 🙏 

Fixes #696

### DIFF
--- a/app/core/layouts/Layout.tsx
+++ b/app/core/layouts/Layout.tsx
@@ -55,7 +55,7 @@ const Layout = ({ title, children, headChildren }: LayoutProps) => {
           </Link>{" "}
           Publishing not possible until resolved.
         </div> */}
-          <div className="flex-grow">{children}</div>
+          <div className="flex flex-grow flex-col">{children}</div>
         </div>
       </RecoilRoot>
       <Footer />
--- a/app/pages/drafts.tsx
+++ b/app/pages/drafts.tsx
@@ -49,12 +49,7 @@ const DraftsContents = ({ expire, signature, currentWorkspace, session, user }) 
 
   return (
     <>
-      <div
-        className="flex w-screen divide-x divide-gray-100 dark:divide-gray-600"
-        style={{
-          height: biggerWindow ? "calc(100vh - 73px - 55px)" : "100%",
-        }}
-      >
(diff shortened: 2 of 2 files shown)


nsunami:
@chartgerink and I chatted a bit— the subtracting height is related to the nav bar and the menu bar (` - 73px - 55px`).

chartgerink (commented):
Whackamole time!

This fix changes the styling of the empty draft state.

Screenshot 2022-10-18 at 09 59 41

If we're okay with that, we could merge. If we're not okay then we need to resolve that first.

Note that a similar empty state (and styling) is used in `/invitations` for what it's worth.

nsunami:
Thanks, Chris!

I think keeping the look consistent makes sense. So, I tried making the draft area grow. 

Then I realized that the Layout file is actually rendering the children as not as a flexbox. So, I updated the Layout file so that the children content is wrapped into a flexbox column.


image

I think it would work all the pages at my glance, but I would appreciate testing from you. 🙏

chartgerink (approved):
It's always risky to make an upstream change to fix something. In this scenario it worked it 😊 

If we can avoid upstream changes, I'd prefer if we do.

PR #746, opened 2022-09-26
Add a confirmation modal when removing a supplementary file (& add an abstracted modal component)

### LINKED ISSUE #695: Add confirmation modal for delete button
When editing a module, it takes one mistaken click to delete a file. If somebody uses R= as their central storage place that might mean important data gets lost.

We could improve the experience by adding a modal asking for confirmation of deleting a single file (similar to when you try to delete the entire module).

Screenshot 2022-09-12 at 14 34 56

nsunami:
This PR adds a confirmation modal when removing a supplementary file. 

image

By implementing this, I abstracted a modal component and created the `Modal.tsx` file. I referred to the existing modal component for deleting a module (`deleteModuleModal.tsx`) for styles.

Fixes #695.

### DIFF
--- a/app/core/components/EditSupportingFileDisplay.tsx
+++ b/app/core/components/EditSupportingFileDisplay.tsx
@@ -4,9 +4,13 @@ import toast from "react-hot-toast"
 import filesize from "filesize"
 
 import deleteSupportingFile from "../../modules/mutations/deleteSupportingFile"
+import { useState } from "react"
+import { Modal } from "../modals/Modal"
 
 const EditSupportingFileDisplay = ({ name, size, url, uuid, moduleId, setQueryData }) => {
   const [deleteMutation] = useMutation(deleteSupportingFile)
+  const [confirmDeleteOpen, setConfirmDeleteOpen] = useState(false)
+
   return (
     <div className="my-2 flex">
       <a
@@ -24,7 +28,23 @@ const EditSupportingFileDisplay = ({ name, size, url, uuid, moduleId, setQueryDa
           <TrashCan
             size={24}
             className="inline-block h-6 w-6 fill-current align-middle text-red-500"
-            onClick={async () => {
+            onClick={() => setConfirmDeleteOpen(true)}
+            aria-label="Delete file"
+          />
+          <Modal
+            title="Confirm removing a supporting file"
+            body={
+              <span>
+                The supporting file, &quot;<span className="font-bold">{name}</span>&quot;, will be
+                removed.
+              </span>
+            }
+            primaryAction="Remove"
+            primaryButtonClass="text-red-700 hover:bg-red-200 focus:ring-red-500 focus:ring-offset-0 focus:ring-2 dark:border dark:border-gray-600 dark:bg-gray-800 dark:text-red-500 dark:hover:border-gray-400 dark:hover:bg-gray-700"
+            isOpen={confirmDeleteOpen}
+            setIsOpen={setConfirmDeleteOpen}
+            onSubmit={async () => {
+              console.log("Deleting")
               toast.promise(deleteMutation({ id: moduleId, uuid }), {
                 loading: "Deleting...",
                 success: (data) => {
@@ -34,7 +54,6 @@ const EditSupportingFileDisplay = ({ name, size, url, uuid, moduleId, setQueryDa
                 error: "Something went wrong...",
               })
             }}
-            aria-label="Delete file"
           />
         </button>
         <a href={url} target="_blank" download rel="noreferrer">
--- a/app/core/components/EditSupportingFileDisplay.tsx
+++ b/app/core/components/EditSupportingFileDisplay.tsx
@@ -35,11 +35,12 @@ const EditSupportingFileDisplay = ({ name, size, url, uuid, moduleId, setQueryDa
             title="Confirm deletion"
             body={
               <span>
-            Upon confirmation, the supporting file &quot;{name}&quot; will be permanently deleted. Are you sure you want to delete this file?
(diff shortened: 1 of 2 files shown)


app/core/components/EditSupportingFileDisplay.tsx
```diff
@@ -24,7 +28,23 @@ const EditSupportingFileDisplay = ({ name, size, url, uuid, moduleId, setQueryDa
...
+                The supporting file, &quot;<span className="font-bold">{name}</span>&quot;, will be
+                removed.
+              </span>
+            }
+            primaryAction="Remove"
+            primaryButtonClass="text-red-700 hover:bg-red-200 focus:ring-red-500 focus:ring-offset-0 focus:ring-2 dark:border dark:border-gray-600 dark:bg-gray-800 dark:text-red-500 dark:hover:border-gray-400 dark:hover:bg-gray-700"
+            isOpen={confirmDeleteOpen}
+            setIsOpen={setConfirmDeleteOpen}
+            onSubmit={async () => {
+              console.log("Deleting")
```
chartgerink, line 47:
```suggestion
```

app/core/components/EditSupportingFileDisplay.tsx
```diff
@@ -24,7 +28,23 @@ const EditSupportingFileDisplay = ({ name, size, url, uuid, moduleId, setQueryDa
...
+          />
+          <Modal
+            title="Confirm removing a supporting file"
+            body={
+              <span>
+                The supporting file, &quot;<span className="font-bold">{name}</span>&quot;, will be
+                removed.
+              </span>
+            }
+            primaryAction="Remove"
```
chartgerink, line 42:
```suggestion
            primaryAction="Delete File"
```

It is good to have the button describe the action. This is for accessibility purposes - clarity comes from being verbose and not implying anything.

app/core/components/EditSupportingFileDisplay.tsx
```diff
@@ -24,7 +28,23 @@ const EditSupportingFileDisplay = ({ name, size, url, uuid, moduleId, setQueryDa
           <TrashCan
             size={24}
             className="inline-block h-6 w-6 fill-current align-middle text-red-500"
-            onClick={async () => {
+            onClick={() => setConfirmDeleteOpen(true)}
+            aria-label="Delete file"
+          />
+          <Modal
+            title="Confirm removing a supporting file"
```
chartgerink, line 35:
```suggestion
            title="Confirm deletion"
```

Consistency in copy is key - here you use `delete` and `remove` interchangeably.

app/core/components/EditSupportingFileDisplay.tsx
```diff
@@ -24,7 +28,23 @@ const EditSupportingFileDisplay = ({ name, size, url, uuid, moduleId, setQueryDa
...
+            body={
+              <span>
+                The supporting file, &quot;<span className="font-bold">{name}</span>&quot;, will be
+                removed.
```
chartgerink, line 39:
```suggestion
            Upon confirmation, the supporting file &quot;{name}&quot; will be permanently deleted. Are you sure you want to delete this file?
```

app/core/modals/Modal.tsx
```diff
@@ -0,0 +1,91 @@
...
+    isOpen,
+    setIsOpen,
+    title,
+    body,
+    onSubmit,
+    primaryAction = "Ok",
+    primaryButtonClass,
+    secondaryAction = "Cancel",
+    secondaryButtonClass,
+  } = props
```
chartgerink, line 15:
```suggestion
export const Modal = ({
    isOpen,
    setIsOpen,
    title,
    body,
    onSubmit,
    primaryAction = "Ok",
    primaryButtonClass,
    secondaryAction = "Cancel",
    secondaryButtonClass,
  }) => {
```

This is a bit confusing - if I now use a prop that isn't required the `Modal` doesn't throw (I think). I would bring the props into the component explicitly.

chartgerink (changes requested):
Left some comments for your review - it's great you abstracted this out - I created a lot of modals and this will help make the code more consistent elsewhere in a future PR (see especially the `modals` folder 😅).

One round of revisions would be enough I think - re-request a review whenever you are ready! 😊

app/core/modals/Modal.tsx
nsunami: Thanks Chris!

The suggested change throws a type error since the parameter's type is expecting any but not undefined.


```
Property 'secondaryButtonClass' is missing in type '{ title: string; body: Element; primaryAction: string; primaryButtonClass: string; isOpen: boolean; setIsOpen: Dispatch<SetStateAction<boolean>>; onSubmit: () => Promise<...>; }' but required in type '{ isOpen: any; setIsOpen: any; title: any; body: any; onSubmit: any; primaryAction?: string | undefined; primaryButtonClass: any; secondaryAction?: string | undefined; secondaryButtonClass: any; }'.ts(2741)
```

@chartgerink Do you think it's a good idea to develop type definitions for these parameters?

chartgerink: Yes it would be good to do Type definitions so that the properties of the function are defined. I didn't test the suggestion so we can go two routes:
1. Fix the issue you raised and then request another review
2. Open an issue so that we bring the props into the component explicitly at a later, but soon, time. We keep as is for now.

Let me know which option you prefer. 😊

nsunami: Thanks for your ideas, Chris! 🙏 

I tried fixing the error by defining the prop types in this PR, using the `prop-types` approach. Would that work for you?

chartgerink (approved):
I made only a slight adjustment to the button styling (as discussed we need to check the consistency throughout the app) - beyond that looks good to me!

PR #756, opened 2022-09-27
Make the "Got it!" text clickable in the email onboarding quest

### LINKED ISSUE #738: Cannot click "Got it" link after agreeing to receive email notifications  🐛
### URL of the Page

https://www.researchequals.com/dashboard

### What happened?

I noticed that after clicking yes 👍 to receive email notification, I cannot click "Got it!" button on the confirmation (as below). 

image

### Steps to reproduce

1. Go to dashboard with an account that has not completed the onboarding quest for email notifications
2. Select "Yes" to receive email notifications
3. Cannot click "Got it!" (I expected that we can close the card by clicking this.)

Chrome

### Anything else?

_No response_

nsunami:
This PR makes the "Got it!" text clickable in the email onboarding quest. Clicking the link will hide the quest card.  Fixes #738.

This PR uses a stateful value (`showCard`) to track the state of the card (show/hide).  This PR also converts the ternary operator style into the short-circuit style, as mentioned in #546.

### DIFF
--- a/app/core/components/OnboardingQuests.jsx
+++ b/app/core/components/OnboardingQuests.jsx
@@ -1,6 +1,6 @@
 import { useMutation, useQuery, Link } from "blitz"
 import { Widget } from "@uploadcare/react-widget"
-import { useRef } from "react"
+import { useRef, useState } from "react"
 import { Formik, Form } from "formik"
 import {
   Email,
@@ -112,10 +112,13 @@ const OnboardingEmail = ({ data }) => {
 const OnboardingEmailAccept = ({ data }) => {
   const [emailConsentMutation, { isSuccess }] = useMutation(changeEmailConsent)
   const [emailNotifications, setEmailNotifications] = useRecoilState(emailNotificationsAtom)
+  const [showCard, setShowCard] = useState(
+    data.emailConsent === null || emailNotifications.emailConsent === null
+  )
 
   return (
     <>
-      {data.emailConsent === null || emailNotifications.emailConsent === null ? (
+      {showCard && (
         <div className="onboarding my-2 flex w-full flex-col rounded-r border-l-4 border-cyan-400 bg-cyan-50 p-4 dark:border-cyan-200 dark:bg-cyan-900 lg:my-0">
           <div className="flex flex-grow">
             <div className="">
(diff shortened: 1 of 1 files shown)


app/core/components/OnboardingQuests.jsx
```diff
@@ -300,11 +308,7 @@ const OnboardingAvatar = ({ data, expire, signature, refetch }) => {
         >
           <div className="flex flex-grow">
             <div className="">
-              <UserAvatar
-                size={32}
-                className="h-5 w-5 text-indigo-400 dark:text-indigo-200"
-                aria-hidden="true"
-              />
+              <UserAvatar size={32} className="h-5 w-5 text-indigo-400 dark:text-indigo-200" />
```
chartgerink, line 311: Could you please tell me why this is edited? This seems like an arbitrary edit and out of scope for this PR. 

It is good practice to not make edits unrelated to the scope as it is high-risk for creating problems/confusion. 

for what it's worth - `aria-hidden` is used to remove icons (among other things) from screenreaders. This removal needs to be undone because it's necessary here.

chartgerink (changes requested):

app/core/components/OnboardingQuests.jsx
chartgerink:
```suggestion
              <UserAvatar size={32} className="h-5 w-5 text-indigo-400 dark:text-indigo-200" aria-hidden="true" />
```

app/core/components/OnboardingQuests.jsx
```diff
@@ -138,7 +141,14 @@ const OnboardingEmailAccept = ({ data }) => {
...
               {isSuccess ? (
-                <p className="whitespace-nowrap font-medium  underline">Got it!</p>
+                <p className="whitespace-nowrap font-medium underline">
+                  <a
+                    className="hover:cursor-pointer hover:text-blue-600"
+                    onClick={() => setShowCard(false)}
+                  >
+                    Got it!
+                  </a>
+                </p>
```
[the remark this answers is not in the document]
nsunami: Got it! I think for me an underlined text implies an action/a-tag visually, but in the code that makes sense. 👍

app/core/components/OnboardingQuests.jsx
nsunami: @chartgerink I don't remember actively removing this. Thanks for adding this back.

I 100% agree with not making arbitrary edits! Thanks for catching.

And thanks for explaining what `aria-hidden` does.

chartgerink (approved):

Issue #721, opened 2022-09-19
Failing to test Stripe on local


nsunami:
I'm exporting the issue here from the PR to preserve the knowledge.

Currently, testing Stripe fails on my local, since the module will not get published. (Even though the Stripe payment succeeds.)

It looks like on my end that `stripe_webhook.ts` may not be firing since the module that I'm trying to publish will continue to have the `published` value as `false`. 

I have not yet figured out how `stripe_webhook.ts` is called in the app.

**From the PR:**
------
I've tried using the ResearchEqual's testing secret, but the publishing did not work on my local.

Steps:

1. Run `stripe listen --forward-to localhost:3000/api/stripe_webhook`
2. Add the webhook signing secret to the env files 
3. Click the publish button
4. Enter the payment info on stripe website. The payment succeeds
5. Module not found error in `[suffix].tsx`

In `[suffix].tsx`, we are looking for a module where `published` is true. But, the module that I'm trying to publish still has its `published` value `false`. I am guessing that the `stripe_webook.ts` is not working correctly on my local.

_Originally posted by @nsunami in https://github.com/libscie/ResearchEquals.com/issues/706#issuecomment-1251415669_

nsunami:
I also opened a related issue where we cannot test Stripe on Heroku: https://github.com/libscie/ResearchEquals.com/issues/736

nsunami:
Shoutout to @chartgerink for the detailed notes on this previous PR! https://github.com/libscie/ResearchEquals.com/pull/161

nsunami:
I was able to set up my own Stripe testing environment on local! 🎉 

How I did:
1. I manually added the product on my Stripe account (e.g., "Collaborative collection")
2. Get the API ID (`price_***`)
3. Update the API ID in the seeding script (`seeds.ts`). In my case, I update this line:

 https://github.com/libscie/ResearchEquals.com/blob/2439a2245e2f05e8c1b2b46b40cbf1d996ca2a83/db/seeds.ts#L454

4. Seed the database (`blitz db seed`)
5. Run the Stripe CLI (`stripe listen --forward-to localhost:3000/api/stripe_webhook`)
6. Match the Stripe secrets with my account (`STRIPE_SECRET_KEY`, `STRIPE_PUBLISHABLE_KEY`, and `STRIPE_WEBHOOK_SECRET`). `STRIPE_WEBHOOK_SECRET` is shown in the console when Stripe CLI is running. 
[...]

PR #763, opened 2022-09-28
Use a state for the editor role to sync the dropdown with what's actually selected

### LINKED ISSUE #742: Users can select collections roles that are not allowed in the drop-down 🐛
### URL of the Page

https://www.researchequals.com/collections/waca-fd/admin

### What happened?

Currently, the user can select a role that's not allowed in the collections' admin panel in the drop-down menu. (Just visually)

For example, in my personal collection's admin page, I can select myself as a USER.

image

It's not allowed, and thus it will throw a snack bar at the bottom with an error message:

image

But then, the selection in the drop-down is now changed to USER. Ideally, we want to fix it to OWNER in this case.

image

### Steps to reproduce

1. Go to your personal collections' admin panel
2. Try to change your role to USER

Chrome

### Anything else?

This issue came up during the review for the collections PR https://github.com/libscie/ResearchEquals.com/pull/685#discussion_r977799858

nsunami:
This PR uses a state for the editor role, so that the selection would correspond to what's allowed in the backend. Fixes #742.

(I'm making this PR a draft since I'm yet to test the collaborative and community collection. Researching about Stripe.)


nsunami:
Now that I can test Stripe on local (#721), I can test for the other types of collections.

nsunami:
While testing this, I realized that I could not change the collections roles. https://github.com/libscie/ResearchEquals.com/issues/765 
And thus I cannot proceed with testing. I think it would make sense to address #765 first before merging this PR.

nsunami:
Now that #765 is closed, I'm making this PR ready for review! (If merge with main is needed, we can do that too, since this branch may be behind the main a bit...)

chartgerink:
@nsunami it looks like there are no conflicts - would you like me to review?

nsunami:
@chartgerink Yes please! Thanks for checking 🙏

chartgerink (approved):
I know this one has been the cause of some friction and stress, and I'm glad we didn't rush this tuesday - that was a good decision of you. 🙌 

Thanks for pulling this along - the pleasure is yours to merge this 😊

PR #767, opened 2022-09-29
Fix the editor roles change & require at least one owner to collections 

### LINKED ISSUE #765: Cannot change user roles in community collection 🐛
### URL of the Page

researchequals.com/collections/[suffix]/admin

### What happened?

I created a community collection on local, and I added another account to the collection.

When I try to change the other person's role, I cannot do so:

https://user-images.githubusercontent.com/17035406/193060723-be992d01-9389-4c0b-8ad3-7b7efa6cc160.mov

### Steps to reproduce

1. Create a community collection
2. Add another user as an editor
3. Change the role from USER to ADMIN

Chrome

### Anything else?

~~While working on this, I noticed that the mutation `changeEditorRole` is not receiving collection information. (I thought we need to pass the ID for the collection to specify which collection's role we are changing). I may be missing something, though.~~
Correction: The `id` in the `Editorship` table (`editorId`) has an implicit relation with the workspace x collection, and thus we don't need to pass collection id (the `editorId` implies which collection we are referring to).

nsunami:
This PR allows the editor roles to change, while requiring a collection to have at least one owner. Fixes #765.

While working on this PR, I was making the assumptions below.
@chartgerink Please point out if any of these are incorrect!

1. A collection should have at least one owner 
2. A collection can have multiple owners
3. A collection can have multiple admins

### DIFF
--- a/app/collections/mutations/changeEditorRole.ts
+++ b/app/collections/mutations/changeEditorRole.ts
@@ -23,6 +23,9 @@ export default resolver.pipe(
         ownerAdmins += 1
       }
     }
+    // Predict the number of owners and admins after the requested change
+    if (role === "OWNER" || role === "ADMIN") ownerAdmins += 1
+
     if (ownerAdmins === 1) throw new Error("Cannot change your role as last admin or owner.")
 
     const editorship = await db.editorship.update({
--- a/app/collections/mutations/changeEditorRole.ts
+++ b/app/collections/mutations/changeEditorRole.ts
@@ -17,9 +17,11 @@ export default resolver.pipe(
       },
     })
 
+    // Set up the counter for the number of owners and admin users
     let ownerAdmins = 0
     let owners = 0
 
+    // Count the total number of existing owners and admin members
     for (const element of oldEditorship?.collection.editors!) {
       if (element.role === "OWNER" || element.role === "ADMIN") {
         ownerAdmins += 1
@@ -34,6 +36,7 @@ export default resolver.pipe(
       if (e.id === editorId && e.role === "OWNER" && role === "ADMIN") owners -= 1
     })
 
+    // Throw an error when changing the last admin or owner
     if (ownerAdmins === 1) throw new Error("Cannot change your role as last admin or owner.")
 
     // Throw an error if there will be no owners after the change
--- a/app/collections/mutations/changeEditorRole.ts
+++ b/app/collections/mutations/changeEditorRole.ts
@@ -18,15 +18,26 @@ export default resolver.pipe(
     })
 
     let ownerAdmins = 0
+    let owners = 0
+
     for (const element of oldEditorship?.collection.editors!) {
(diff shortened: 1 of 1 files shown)


nsunami:
Tested on my local:

For both collaborative & community collections
1. Cannot change as the sole owner of the collection (at least one owner required) ✅ 
2. When I add another person, I can grant ownership to the person ✅  then I can be an admin. ✅
3. I can assign everyone to be owner ✅
4. I cannot assign everyone to be an admin, since at least one owner is required ✅

app/collections/mutations/changeEditorRole.ts
```diff
@@ -17,14 +17,31 @@ export default resolver.pipe(
...
+
+    // Count the total number of existing owners and admin members
     for (const element of oldEditorship?.collection.editors!) {
       if (element.role === "OWNER" || element.role === "ADMIN") {
         ownerAdmins += 1
       }
+      if (element.role === "OWNER") owners += 1
     }
+    // Predict the number of owners and admins after the requested change
+    if (role === "OWNER" || role === "ADMIN") ownerAdmins += 1
```
chartgerink, line 32:
```suggestion
    if ((role === "OWNER" || role === "ADMIN") && oldEditorship.role === "USER") ownerAdmins += 1
```

I am trying to follow this logic, but struggling without access to a test environment. It seems like we might need to include a check that the editor is in fact not already an `OWNER || ADMIN`, such that we don't overcount the `ownerAdmins`?

For example, we might end in the situation where somebody tries to change to USER as last OWNER - gets an error, and then changes back to OWNER. Here we might be double counting the amount of `ownerAdmins` and not hitting the condition of `ownerAdmins === 1`.

What do you think @nsunami?

chartgerink (changes requested):
I think we're pretty much there but I want to go through a round of revisions to be safe. Feel free to re-request a review/ping me about what you think! 🥜

app/collections/mutations/changeEditorRole.ts
nsunami: Thanks for the catch, Chris! 

While testing this, I found a problem, which is related to the situation that you describe. 

1. Set up a collection with at least another user:
image

2. Change the other person's role as an owner
image

3. Set my role as an admin
image

4. Set the other person's role as a user
image

5. Now we have a collection without an owner, which should not be the case.

I will need to think about this.

nsunami:
@chartgerink I ended up overhauling the logic, and I think this may be a better direction.

1. I created `futureEditors` array to predict future state of the editors array
2. Then, I count the numbers for the owners
3. If there will be no owner (0) after the requested change, throw an error.

I ended up removing the logic that counted the number of existing owners and admins. Together with the error throwing logic. I did so because the current code already predicts the number of owners. By doing so, we assure that there will be at least one owner.

chartgerink (approved):
Nicely done @nsunami 🙏 The improvements are reliable - thanks for pushing this along.

Issue #781, opened 2022-10-04
Markdown files are coming back as a binary from Uploadcare (not `text/markdown`)  🐛 


nsunami:
### URL of the Page

https://www.researchequals.com/drafts

### What happened?

When getting the uploaded `.md` file from Uploadcare, the MIME type comes out as binary (application/octet-stream). 

<img width="1127" alt="image" src="https://user-images.githubusercontent.com/17035406/193868939-95a8efbf-4e09-4efa-8774-70b54adca806.png">

This is problematic for rendering the preview, since we rely on the file's MIME type being `text/markdown`. 

Subsequently, the file is not previewed:

<img width="682" alt="image" src="https://user-images.githubusercontent.com/17035406/193869410-5462c972-f8b6-4ed5-9c27-36232e91cc7b.png">

### Steps to reproduce

1. Create a draft
2. Upload a`.md` file as the main file

### What browser(s) did you use?

Chrome

### Anything else?

I can see that the published module on production shows the markdown file.

For example, [this module](https://www.researchequals.com/modules/spj3-x65) previews the markdown content just fine. 

<img width="934" alt="image" src="https://user-images.githubusercontent.com/17035406/193870151-c308d2a1-89d1-4932-b2d3-9bd8fc65d566.png">

But, in my local dev environment, even published modules are not previewing:
<img width="916" alt="image" src="https://user-images.githubusercontent.com/17035406/193870298-c5e77928-750b-42a6-9171-2569bbe9af93.png">

nsunami:
It would be helpful to see if the file uploaded to Uploadcare is registered as markdown or binary on their end.

chartgerink:
Uploadcare has this file registered as `application/octet-stream` and others as `text/markdown` (those that preview). 

What OS are you running? For me it all works when uploading so I cannot reproduce at the moment. In essence the preview functionality is working but it looks like the mimetype is not properly registered upon upload (hence retrieval).

Note: I downloaded your `test-markdown.md` from Uploadcare and uploaded it - it did register `text/markdown` at that point.

nsunami:
Thanks for checking, @chartgerink!

I was using MacOS 12.6

I checked it on my Chrome, and and also with my incognito mode on the upload, but it did not resolve the issue... not sure why



https://user-images.githubusercontent.com/17035406/197219552-548bdf13-6f35-4e25-aa99-d1256be8c172.mov

[...]

PR #782, opened 2022-10-04
Convert ternary operators to short-circuits in the `FilePreviewer` component

### LINKED ISSUE #754: Audit the ternary operators in the `FilePreviewer.tsx` file and convert them to short-circuit, if possible
Breaking down #546: I saw that the `FilePreviwer.tsx` has some ternary operators that may be converted to short-circuit syntax. Creating this issue to track the conversion.

nsunami:
This PR converts ternary operators to short-circuits in the `FilePreviewer` component. Fixes #754.

I'm marking this PR as a draft, since I found that Markdown files are not previewing during the testing. (See Issue #781)
I doubt that this PR introduces this issue, but I'd appreciate another person's look on this. 

Files checked:

1. Image ✅ 
2. Office file ✅ 
3. Markdown 🚫  (Issue #781)

### DIFF
--- a/app/core/components/FilePreviewer.tsx
+++ b/app/core/components/FilePreviewer.tsx
@@ -29,6 +29,7 @@ const MainFileViewer = ({ mainFile }) => {
           {/* Preview image */}
           {mainFile.isImage && (
             <img
+              alt="An image of the main file"
               src={mainFile.cdnUrl}
               className="mx-auto my-2 h-auto w-full"
             />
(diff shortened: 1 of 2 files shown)


chartgerink:
@nsunami - is this still actual? There's also a conflict right now that would need to be resolved.

PS: I see an argument is removed from the component - that seems outside the scope of the described PR?

nsunami:
@chartgerink thanks for the heads-up!

The argument was removed because that was not used. In my head, it fell in the broader goal of this issue to clean up this component (but I admit that I stretched the goal).

Do you prefer scoping this out to another issue? I'm open 👍 

Also, working on to resolving the conflict!

chartgerink (approved):
Looking good!

Love it when a PR is so easy to review 🙏

PR #793, opened 2022-10-06
Use the `<Image>` component instead of `<img>` element in `FilePreviewer`

### LINKED ISSUE #776: `no-img-element` Warning in the file previewer (`FilePreviewer.tsx`) 🐛
### URL of the Page

drafts

### What happened?

I noticed that we have `` tag used in the file previewer component (`FilePreviewer.tsx`). Next.js throws a warning if we are not using `<Image>` component ([Next.js docs](https://nextjs.org/docs/messages/no-img-element)).

Ideally, we want to use `<Image>` component. To do so, we need to specify width and height of the image beforehand. ~~(not sure how to get that info.)~~

Correction: it looks like the file info object we get from the CDN has `originalImageInfo` → `width` & `height`. So probably, it will be straightforward to use those. 

### Steps to reproduce

1. `npm run dev`
2. Make a draft
3. Upload an image as the main file

Chrome

### Anything else?

_No response_

nsunami:
This PR replaces the `` element in the `FilePreviewer` component, with the `<Image>` component. Doing so will suppress the `no-img-element` warning in Next.js. 

This PR also adds the Uploadcare CDN to the list of allowed domains in the Blitz config file (`blitz.config.ts`). 

Fixes #776.


chartgerink (changes requested):
In essence I think this is a great improvement. I'd been putting off doing this because there's so many places where this is relevant - this gets us on the road there 🙏 

At the moment, we do get an error if we implement this for the test environment - the images we generate in the seed script don't have a width or height. I suggested a quick fix but that needs to be tested.

Screenshot 2022-10-18 at 10 59 02

It is important we ensure the test environment doesn't have arbitrary errors so I would like to fix this first. Then we can move forward. As a note - it might be an idea to think about doing this for all `` tags throughout the platform.

nsunami:
There are still two things (at least) that I will need to work on here!

1. Make the width and height props optional 
```
              width={mainFile?.originalImageInfo?.width || 300}
              height={mainFile?.originalImageInfo?.height || 300}
```

2. I'm getting an error for cross-site image rendering. The test images are coming from loremflickr.com, and I'm told to add that to our blitz.config.ts file.

```
    domains: ["eu.ui-avatars.com", "ucarecdn.com", "loremflickr.com"],

```

chartgerink:
Oki - re-request whenever you're ready. I would suggest to not allow loremflickr.com but adjust the procedure where it comes from. We don't know this domain so it's a potential issue down the line.

chartgerink:
Any update on this?

nsunami:
It happens Faker pulls cat images from `loremflickr.com` when using [`faker.image.cats()`](https://github.com/faker-js/faker#:~:text=Image-,faker.image.cats(),-https%3A//loremflickr.com), which is the root of the problem 🐱 

I'm thinking about two potential approaches to solve this:

1. Find a way to have test images from `ucarecdn.com` (or fix it to some file uploaded to uploadcare)
2. Allow `loremflickr.com` only in the testing environment, but not in production

Probably Option 1 is better?

nsunami:
Following Option 1 I mentioned earlier, I drafted out a revision for our seeding script. 

Basically, my idea is to use a static image hosted on uploadcare, and use that as an image across the fake entries.

For example, if we want to set the cat image as [this one](https://ucarecdn.com/a91fd01f-e5e8-4d07-b4b1-1896dc2742f4/-/preview/1162x693/-/setfill/ffffff/-/format/jpeg/-/progressive/yes/), we would set our our seeds (`/Users/nami/git/ResearchEquals.com/db/seeds.ts`) as ..

```js
...
      const catImage = `https://ucarecdn.com/a91fd01f-e5e8-4d07-b4b1-1896dc2742f4/-/preview/1162x693/-/setfill/ffffff/-/format/jpeg/-/progressive/yes/`
      module = await db.module.create({
        data: {
          prefix: "10.53962",
          suffix: await generateSuffix(undefined),
          title: faker.lorem.sentence(),
          description: faker.lorem.sentences(5 * (index + 1)),
          published: true,
          publishedWhere: "ResearchEquals",
          publishedAt: faker.date.past(),
          url: `https://doi.org/10.53962/${suffix}`,
          main: {
            name: faker.system.fileName(),
            size: 603268,
            uuid: faker.datatype.uuid(),
            cdnUrl: catImage,
            isImage: true,
            isStored: true,
            mimeType: "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
            sourceInfo: { file: {}, source: "local" },
            originalUrl: catImage,
            cdnUrlModifiers: null,
            originalImageInfo: null,
          },
...
```

chartgerink:
Option 1 indeed sounds good - is the ucarecdn one that we control? (i.e., where did you get the url from?) Would like to ensure that we don't get an arbitrary error if somebody ends up deleting the file and we don't know about it.

I also just saw the mimetype is an office file, could we adjust it to the image? Would prevent it trying to render the image as an office file :)

nsunami:
Got it 👍 

I uploaded the image to R= uploadcare. The URL is here:

https://ucarecdn.com/fea4c143-6382-467d-a500-23826acbb1e4/

The MIME type should be `image/jpeg` now. Would that work?

chartgerink (approved):
Looking good @nsunami!

Note to us is to potentially add self declaring of the alt description when people upload an image as main file. This alt is not going to be enough for people using screenreaders.

nsunami:
> Note to us is to potentially add self declaring of the alt description when people upload an image as main file. This alt is not going to be enough for people using screenreaders.

Great idea 💡 I created an issue so that we won't forget: https://github.com/libscie/ResearchEquals.com/issues/901

nsunami:
@chartgerink I updated the seeds, and re-requesting your review!

chartgerink (approved):

PR #794, opened 2022-10-06
Prevent a "USER" account from publishing a collection

### LINKED ISSUE #769: A user of a collection can make the collection public
I noticed in my testing that a user of a collection can make the collection public:

https://user-images.githubusercontent.com/17035406/193126121-ee85891b-84ed-427e-8723-b27fd806c24b.mov

I expected that only admin or owner can change the visibility—but, I'm not 100% sure about the allowed actions for different user roles. (and thus I'm creating this issue as a question, not as a bug).

nsunami:
This PR prevents a user account with "USER" role from publishing the collection. 

Specifically, this PR hides the button to make the collection public. The banner still shows up:

image

Fixes #769


chartgerink (approved):
Nice catch! 🙏

PR #800, opened 2022-10-07
Use named imports for `filesize` 

### LINKED ISSUE #799: Build fails with the update to `filesize`  🐛
### URL of the Page

_No response_

### What happened?

The dependabot upgrade to `filesize` fails to build the app (https://github.com/libscie/ResearchEquals.com/pull/771). The upgrade requires explicit, named exports from `"filesize"` other than defaults.

Currently, we have:

```js
import filesize from "filesize"
```

We need to update into: 

```js
import { filesize } from "filesize"
```

## Log

```
Failed to compile.

./app/core/components/EditMainFileDisplay.tsx:21:16
Type error: This expression is not callable.
  Type 'typeof import("/Users/nami/git/ResearchEquals.com/node_modules/filesize/types/filesize")' has no call signatures.

  19 |         <span className="inline-block  flex-grow align-middle">{name}</span>
  20 |         {/* https://www.npmjs.com/package/filesize */}
> 21 |         <span>{filesize(size)}</span>
     |                ^
  22 |       </a>
  23 |       <p className="flex">
  24 |         <button className="mx-2">
info  - Checking validity of types .%                                                

```

### Steps to reproduce

`npm run buiild` 

Chrome

### Anything else?

_No response_

nsunami:
This PR converts the default imports for `filesize` into the named imports, fixing the issue introduced by the dependency upgrade (observed in this PR #771 ).

Fixes #799

### DIFF
--- a/app/core/components/EditMainFileDisplay.tsx
+++ b/app/core/components/EditMainFileDisplay.tsx
@@ -1,7 +1,7 @@
 import { Download, TrashCan } from "@carbon/icons-react"
 import { useMutation } from "blitz"
 import toast from "react-hot-toast"
-import filesize from "filesize"
+import { filesize } from "filesize"
 
 import deleteMainFile from "../../modules/mutations/deleteMainFile"
 
--- a/app/core/components/EditSupportingFileDisplay.tsx
+++ b/app/core/components/EditSupportingFileDisplay.tsx
@@ -1,7 +1,7 @@
 import { Download, TrashCan } from "@carbon/icons-react"
 import { useMutation } from "blitz"
 import toast from "react-hot-toast"
-import filesize from "filesize"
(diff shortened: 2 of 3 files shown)


chartgerink (changes requested):
I am closing this pull request as what is happening here is classic code drift. 

You're making changes to `main` based upon changes you know happening in #771. But those changes are not in `main` so the build fails. Yet we cannot merge #771 without having these changes.

Please make these changes directly in #771 so that we can verify the build works and work towards merging 😊

nsunami:
@chartgerink Thanks for pointing that out! And it makes sense that the build fails here because the app build with the older `filesize` package. 

This error got pushed since I was using `npm install --no-save` for checking upgrades, and I forgot to do `npm install` periodically.

PR #816, opened 2022-10-11
Add Zotero metadata for modules

### LINKED ISSUE #626: Add metadata for Zotero ingestion
In order to improve saving ResearchEquals modules to Zotero, we could upgrade the metadata exposed by default in the `<head>` tag. 

For reference, see some of [the documentation from Zotero](https://www.zotero.org/support/dev/exposing_metadata).

This issue tracks the work on this, and potential questions that come up while implementing this. Sub-issues may be added as they come up.

### Resources

-  [Zotero translators](https://www.zotero.org/support/translators) (maybe relevant?)

nsunami:
This PR adds Zotero metadata for modules. I see this as an experimental PR since we may need to tweak what information we want to add, as we discuss in #626. Nevertheless, I wanted to open a PR to show what I got so far. 

Fixes #626.

### DIFF
--- a/app/pages/modules/[suffix].tsx
+++ b/app/pages/modules/[suffix].tsx
@@ -116,7 +116,6 @@ const Module = ({ module, mainFile, supportingRaw }) => {
     arrowColor = "transparent"
   }
 
-  console.log(mainFile)
   useEffect(() => {
     if (mainFile.mimeType === "text/markdown") {
       fetch(mainFile.cdnUrl)


app/pages/modules/[suffix].tsx
```diff
@@ -116,6 +116,7 @@ const Module = ({ module, mainFile, supportingRaw }) => {
     arrowColor = "transparent"
   }
 
+  console.log(mainFile)
```
Comment (chartgerink) [inline app/pages/modules/[suffix].tsx:119]:
```suggestion

```

chartgerink (approved):
Looks great @nsunami! 🙏 

I changed a few things around but the core looks great. Changes: 

- Removed the comments (tags are super clear on themselves)
- Added the `journal` tag
- Added the `publisher` tag
- Added the `citation_pdf_url` in case the main file is a PDF (so it gets imported into Zotero too - oddly enough it does not accept other files it seems?)

I also tested the import for Zotero directly:

Screenshot 2022-10-19 at 18 48 24

PR #820, opened 2022-10-12
Use 404 as a basis for moving from CrossRef to DataCite when adding a reference

### LINKED ISSUE #787: Catch 404 from CrossRef instead of catching all errors
As I mentioned in #747 and https://github.com/libscie/ResearchEquals.com/pull/785#issuecomment-1268768572, the db mutation for adding a reference module (`createReferenceModule`) catches all errors trying to use CrossRef API, and then resolving to use DataCite.

This logic makes a bit harder to debug, since the console will be flooded with the error message from DataCite. I think it would be better to catch 404 from CrossRef, and then resolving to use DataCite. Any other errors will stop the execution, and be reported in the console accordingly.

nsunami:
This PR revises the logic when adding a reference via DOI. The logic tries CrossRef first, then if the error code was 404, it tries DataCite. If the error was other than 404, it will throw a new error with a message.

Previously, the logic was to use DataCite for all errors. 

Fixes #787

### DIFF
--- a/app/modules/mutations/createReferenceModule.ts
+++ b/app/modules/mutations/createReferenceModule.ts
@@ -83,77 +83,86 @@ export default resolver.pipe(resolver.authorize(), async ({ doi }, ctx) => {
     })
 
     return module
-  } catch {
-    const cr = await axios.get(`https://api.datacite.org/dois/${doi}`)
-    const metadata = cr.data
-
-    const module = await db.module.create({
-      data: {
-        published: true,
-        publishedAt: new Date(
-          metadata.data.attributes.publicationYear.toString() || metadata.data.attributes.published
-        ),
-        publishedWhere: metadata.data.attributes.publisher,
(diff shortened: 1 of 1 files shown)


app/modules/mutations/createReferenceModule.ts
```diff
@@ -10,7 +10,11 @@ function capitalizeFirstLetter(string) {
   return string.charAt(0).toUpperCase() + string.slice(1)
 }
 
-export default resolver.pipe(resolver.authorize(), async ({ doi }, ctx) => {
+interface Input {
+  doi: string
+}
+
```
chartgerink, line 16: I get the improvement itself, but please keep PRs related to the topic (it can be very distracting as I know it does not relate here, which is not always clear).

chartgerink (approved):

chartgerink:
Thanks @nsunami - this is much better 😊

PR #859, opened 2022-10-25
Renew Apple Pay merchant file

chartgerink:
I noticed the Apple Pay integration didn't work and discovered it was 
still linked to the wrong canonical URL - `researchequals.com` instead 
of `www.researchequals.com`...

- remove old merchant id file
- Add new Apple Pay merchant file


chartgerink:
Merging with admin privileges because this is not testable - let me know if you'd like more information on this regardless!

PR #861, opened 2022-10-26
Add admin reminting for collections

chartgerink:
This PR adds reminting for collections in the `/admin` (superadmin) portal.

Screenshot 2022-10-26 at 10 58 22

This allows for straightforward reminting of the DOI metadata such that errors can be remedied, or metadata updated (e.g., author names changed).

Steps taken:

1. Display collections in the admin page
2. Create the `updateCrossRefCollection` mutation
3. Add the functionality of the mutation to the admin page

This is a pattern we've previously setup for reminting the DOI metadata for modules.

Note: The `/admin` page is getting crowded and could benefit from an improvement :) But this isn't user facing so not super urgent right now.

### DIFF
--- a/app/collections/mutations/updateCrossRefCollection.ts
+++ b/app/collections/mutations/updateCrossRefCollection.ts
@@ -1,0 +1,46 @@
+import { resolver } from "blitz"
+import db from "db"
+import { isURI } from "../../core/crossref/ai_program"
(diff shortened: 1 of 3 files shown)


nsunami (approved):
LGTM! ✅ 

And I agree that redesign would benefit the page. (Scrolling to find the section for the collection would be a bit hard if there are many modules and collections. 💡)

PR #862, opened 2022-10-26
Make the quick draft appear above the collections banner & make the collections modal appear above everything

### LINKED ISSUE #847: The quick draft modal is covered by the banner for the collections 🐛
### URL of the Page

https://www.researchequals.com/browse

### What happened?

Currently, the banner for the collection is showing up above the quick draft modal 😢 

image

### Steps to reproduce

1. Go to https://www.researchequals.com/browse
2. Open the quick draft modal by pressing "+ Draft"

Chrome

### Anything else?

I think this UI bug was introduced by https://github.com/libscie/ResearchEquals.com/pull/817, when I was adjusting the z-index of the collections modal. 

I forgot to test whether the quick draft appears ok.

nsunami:
This PR fixes two points related to z-indices:

1. This PR makes the quick draft appear above the collections banner. To do so, I adjusted the z-index of the collections banner itself to `z-10`. I previously set to `z-50`, but in retrospect, that was too high for the banner. 😓  (We still want modals and slide-outs to show up on top, for example 🤞 ).

2. This PR makes the collections modal appear on top, by setting `z-50`. (Fixing the point mentioned by @chartgerink on https://github.com/libscie/ResearchEquals.com/issues/847#issuecomment-1288655909) Previously, the class name was `z-999`, but I think it was not a valid Tailwind class, and thus it was not applied.

(I did not separate the PRs because we identified these issues in #847. I'm open to separating if you think that's better, @chartgerink 👍 )

Fixes #847 ✅

### DIFF
--- a/app/core/components/Navbar.tsx
+++ b/app/core/components/Navbar.tsx
@@ -23,7 +23,7 @@ const Navbar = ({
 }) => {
   return (
     <>
-      <div className="z-50 mx-auto w-full border-b border-gray-100 bg-white px-4 dark:border-gray-600 dark:bg-gray-900 sm:px-6 lg:px-8">
+      <div className="z-10 mx-auto w-full border-b border-gray-100 bg-white px-4 dark:border-gray-600 dark:bg-gray-900 sm:px-6 lg:px-8">
         <div className="relative flex justify-between lg:gap-8 xl:grid xl:grid-cols-12">
           <div className="flex md:absolute md:inset-y-0 md:left-0 lg:static xl:col-span-2">
             <div className="my-2 flex shrink-0 items-center">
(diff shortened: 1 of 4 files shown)


chartgerink (changes requested):
Thanks @nsunami - it is looking good except that the navigation bar is over the quick draft.

image

I think this could be resolved by changing `z-50` to `z-10` for `app/core/components/Navbar` which has `z-50` - I'm not 100% about this so please double-check.

nsunami:
@chartgerink Thanks!

I set the nav bar to `z-10`, then it was no longer covering the quick draft slide out.

But, this introduced a new problem where the nav bar is covering the manage authors slide out (z-index was not set there).

image

So, I ended up changing the z-index for the manage authors to `z-10` as well.  (`app/modules/components/ManageAuthors.tsx`)

image


(It may be nice to refactor, and make a component for slide outs—if we plan using slide outs in other places in the future)

Would this work?

chartgerink (approved):
Looking good!

Agree that we should refactor so the slide outs are more easily managable. Note that this also is used elsewhere (like "Following") so we'd need to have a good index of everything first to prevent issues.

nsunami:
Makes sense! Agreed that we need a list/map of where the slide outs are used. 
Merging this now ✅

PR #865, opened 2022-10-27
Add own collections to `/collections`

### LINKED ISSUE #768: Cannot access one's own collection that's not published yet
I noticed that there's no easy way to view one's own collection that's not published yet.

For example, I can create a collaborative collection, and I can see the admin page:

image

Now, if I don't click "Make public" there, and navigate out from the page, there's no easy way to go back to the admin panel of that collection, aside from accessing the URL. (As far as I can tell.) 

It would be nice to have them listed under (1) the Collections tab or (b) the Drafts tab.

chartgerink:
This PR adds the collections that the workspace is an editor of to the `/collections` page - this means that it includes both public and non-public ones.

Fixes #768.

Note: This is not the best solution in the world. We can always further improve this in the future. 😊 I did not pick up the suggestion there because it would take considerably more work (I expect) to rework the `/drafts` as suggested.

## Screenshots

Screenshot 2022-10-27 at 15 30 02


app/collections/queries/getDraftCollections.ts
```diff
@@ -0,0 +1,47 @@
+import { resolver, Ctx, AuthorizationError } from "blitz"
+import db from "db"
+
+export default resolver.pipe(async ({ session }) => {
+  const draftCollections = await db.collection.findMany({
```
nsunami, line 5: Are we getting all of one's collection, not only the draft collections, in this query? (If so, shall we change to something like `getMyCollections`?)

app/pages/collections.tsx
```diff
@@ -38,6 +38,86 @@ const CollectionsPage: BlitzPage = () => {
...
+                                : collection.submissions.length
+                            }-teal`
+                          : `col-span-1 bg-amber-200 text-base dark:bg-amber-600 md:col-span-1 collection-${
+                              collection.submissions.length > 4
+                                ? "4"
+                                : collection.submissions.length
+                            }-amber`
+                      }  cursor-pointer rounded-md bg-cover bg-center`}
+                      key={`${collection.title}-${index}`}
+                    >
```
nsunami, line 69: I noticed that some of these are repeated later (under all collections), which makes a good case for refactoring. No need to resolve this on this PR—I just wanted to leave a comment 👍

nsunami (changes requested):
I think this is a great patch!  (We don't want users to think that their paid collection is gone!)
Working well on my local 👍 

I had one point to clarify (the name of the query) and one comment here.

Hope that helps!

app/collections/queries/getDraftCollections.ts
chartgerink: Fair point - will update!

app/pages/collections.tsx
chartgerink: 💯%

sloppy of me 😅 will look into refactoring this when I update the filiename

nsunami: No worries! And, if you want me to take over refactoring, I'm happy to do so in another issue–PR 👍

PR #866, opened 2022-10-27
Refactor collections admin to have save buttons

chartgerink:
We are experiencing problems in production with the saving of the title, subtitle, and description based on the `onBlur` save action. It worked in test but that's useless if it doesn't work in prod :) 

So now we adjust to have the admin portal display buttons to actively save - this should resolve the issue (although we can only know for sure when it's in production!).

Screenshot 2022-10-27 at 16 47 01

### DIFF
--- a/app/collections/components/AdminSubtitle.tsx
+++ b/app/collections/components/AdminSubtitle.tsx
@@ -1,54 +1,64 @@
-import { useMutation } from "blitz"
+import { useMutation, validateZodSchema } from "blitz"
 import { Field, Form, Formik } from "formik"
 import toast from "react-hot-toast"
 import changeSubtitle from "../mutations/changeSubtitle"
+import { useFormik } from "formik"
+import { z } from "zod"
 
 const Subtitle = ({ collection, refetchFn, isAdmin }) => {
   const [changeSubtitleMutation, { isSuccess: isSubtitleSuccess }] = useMutation(changeSubtitle)
+
+  const formik = useFormik({
+    initialValues: {
+      subtitle: collection.subtitle,
+    },
+    validate: validateZodSchema(
+      z.object({
+        subtitle: z.string().min(1).max(300),
+      })
+    ),
+    onSubmit: async (values) => {
+      toast.promise(
+        changeSubtitleMutation({
+          id: collection.id,
+          subtitle: values.subtitle,
+        }),
+        {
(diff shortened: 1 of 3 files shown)


app/collections/components/AdminSubtitle.tsx
```diff
@@ -1,54 +1,64 @@
...
-              }}
+            <label htmlFor="subtitle" className="sr-only">
+              subtitle
+            </label>
+            <textarea
+              rows={1}
+              id="subtitle"
+              className="w-full border-0 text-center text-sm font-medium leading-5 focus:ring-0 dark:bg-gray-900 md:text-base "
+              {...formik.getFieldProps("subtitle")}
+            />
```
nsunami, line 54: Just a thought—I wonder if it's useful to show the error here for the zod validation. Would something like this work? (ideally we want to show different errors for the lacking (0) vs too long (300+) titles, but it may be too much overhead for now?)

```suggestion
            />
            {formik.errors.subtitle && (
              <div className="border-none text-center text-sm font-medium text-gray-700 dark:text-gray-100">
                {"The subtitle should be 1 to 300 characters long."}
              </div>
            )}
```

app/collections/components/AdminTitle.tsx
```diff
@@ -1,54 +1,64 @@
...
-              }}
+            <label htmlFor="title" className="sr-only">
+              title
+            </label>
+            <textarea
+              rows={1}
+              id="title"
+              className="w-full select-none overflow-auto border-0 bg-white text-center text-3xl focus:ring-0 dark:bg-gray-900 md:text-5xl lg:text-6xl "
+              {...formik.getFieldProps("title")}
+            />
```
nsunami, line 54: Similarly from the above comment, it would be something like this?


```suggestion
            />
            {formik.errors.title && (
              <div className="border-none text-center text-sm font-medium text-gray-700 dark:text-gray-100">
                {"The title should be 1 to 300 characters long."}
              </div>
            )}
```

nsunami (approved):
I think this works well for the current implementation!

I left suggestions for showing errors for character limits---it's up to you if you include them or not 👍

app/collections/components/AdminTitle.tsx
chartgerink: Great suggestions! Updated.

PR #876, opened 2022-10-31
Improve layering of collections confirmation modal

chartgerink:
Just another layering issue that needs to be fixed as soon as possible.

Confirmation modal for collections was behind the options modal (z-10 confirmation on z-50 collections modal). Now it should be on top again (both z-50).


chartgerink:
Merging because it's priority and seems to work after self-review!

PR #882, opened 2022-11-01
Show a confirmation modal when switching from ADMIN/OWNER to USER

### LINKED ISSUE #766: Confirm when switching one's role from owner/admin to a user
In collections, the owner and admin can change themselves as a user, but doing so immediately loses access to the admin panel for the dashboard.

I think it would be nicer experience to show a modal to confirm that they want to step out from being an owner/admin.

nsunami:
This PR shows a confirmation modal when switching from ADMIN/OWNER to USER. 

Here are things that I want to highlight:

1. modified the `<Modal />` component to allow an action when cancelling out from a modal `onCancel`. When the user clicks cancel, the role will be reverted to the default value (`editor.role`)
2. refactored out the `toast.promise` routine as `onChangeEditorRole`, since it was needed in two places. 
3. improved the state tracking during the mutation error. When the mutation fails, the state value will go back to the default value (`editor.role`). 

Fixes #766


app/collections/components/AdminEditors.tsx
```diff
@@ -60,6 +67,23 @@ const EditorCard = ({ editor, isAdmin, isSelf, refetchFn }) => {
             </select>
             <SetEditorToInactiveModal editor={editor} refetchFn={refetchFn} />
             <DeleteEditorModal editor={editor} refetchFn={refetchFn} />
+            <Modal
+              title="Confirm changing your role to a user"
```
chartgerink, line 71:
```suggestion
              title="Confirm role change"
```

app/collections/components/AdminEditors.tsx
```diff
@@ -60,6 +67,23 @@ const EditorCard = ({ editor, isAdmin, isSelf, refetchFn }) => {
...
+              body={
+                <span>
+                  Are you sure you want to remove yourself as {currentRole.toLowerCase()} of this
+                  collection? By doing so, you will lose access to the admin panel.
```
chartgerink, line 75:
```suggestion
                  Are you sure you want to remove yourself as administrator or owner of this
                  collection? You cannot undo this and will no longer be able to administrate the editors of this collection.
```

chartgerink: The `currentRole` was user in the displayed box.

app/collections/components/AdminEditors.tsx
```diff
@@ -60,6 +67,23 @@ const EditorCard = ({ editor, isAdmin, isSelf, refetchFn }) => {
...
+                </span>
+              }
+              primaryAction="Assign yourself as a user"
```
chartgerink, line 78:
```suggestion
              primaryAction="Change Role"
```

This is more consistent with the rest of the copy.

app/core/modals/Modal.tsx
```diff
@@ -8,6 +8,7 @@ export const Modal = ({
   title,
   body,
   onSubmit,
+  onCancel,
```
chartgerink, line 11: This addition breaks the use of the modal component elsewhere.

Whenever you change an upstream function that is used elsewhere, it needs to be tested in all downstream instances. When you make a change like this, I would recommend checking where else this component/function is used to see whether it works there.

chartgerink (changes requested):
This looks good overall, but the changes to the upstream Modal component are breaking its use elsewhere.

app/core/modals/Modal.tsx
nsunami: Thanks for the catch! I'm updating the Modal, so that `onCancel` is an empty function. It looks like it's working on the draft page.

app/core/modals/Modal.tsx
```diff
@@ -8,6 +8,9 @@ export const Modal = ({
   title,
   body,
   onSubmit,
+  onCancel = () => {
+    null
+  },
```
chartgerink, line 13: Thanks for fixing this 🙏

chartgerink (approved):
Works like a charm 📈

PR #891, opened 2022-11-02
Remove `react-tooltip` as a dependency

### LINKED ISSUE #888: Remove `react-tooltip` as a dependency
With the marketing page update (#870), we are no longer using the `react-tooltip`. We can remove it from the dependency list ✅ 🪶

nsunami:
This PR removes `react-tooltip` as a dependency, since we are no longer using it after the marketing page port. 

Fixes #888 

Also....
giphy


chartgerink (approved):
Perfect, thanks. Also removing from the wiki in 3-2-1... 😊

PR #913, opened 2022-11-10
Add classification algorithm to module types

chartgerink:
This PR adds the classification algorithm as module type, requested by someone. 

In reviewing, we might want to test the display of it because it's a relatively lenthy module type.

### DIFF
--- a/db/seeds.ts
+++ b/db/seeds.ts
@@ -50,7 +50,7 @@ const seed = async () => {
     { wikidata: "Q429785", name: "Poster", schema: "Poster" },
     { wikidata: "Q59094171", name: "Ethical review", schema: "CreativeWork" },
     { wikidata: "Q580922", name: "Preprint", schema: "CreativeWork" },
-    { wikidata: "Q1744627", name: "Classification Algorithm", schema: "CreativeWork" },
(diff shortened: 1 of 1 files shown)


chartgerink:
@all-contributors add @marioangst for ideas

db/seeds.ts
```diff
@@ -50,6 +50,7 @@ const seed = async () => {
     { wikidata: "Q429785", name: "Poster", schema: "Poster" },
     { wikidata: "Q59094171", name: "Ethical review", schema: "CreativeWork" },
     { wikidata: "Q580922", name: "Preprint", schema: "CreativeWork" },
+    { wikidata: "Q1744627", name: "Classification Algorithm", schema: "CreativeWork" },
```
nsunami, line 53: I noticed that existing names use sentence case (e.g., "Ethical review") while this change uses the title case. Is this intended? 

In case we want to keep title case:

```suggestion
    { wikidata: "Q1744627", name: "Classification algorithm", schema: "CreativeWork" },
```

nsunami (changes requested):
@chartgerink Nice addition! I had one small comment above that I'd want to double-check with you 👍

nsunami (approved):
The type is rendering good in the module view ✅ 

image

image

PR #923, opened 2022-11-14
Update dependabot.yml

chartgerink:
Update dependabot to run once a week on Wednesday, as @nsunami, our main dependabot'er requested 😊 

Based this on the [docs for dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,8 @@ updates:
     schedule:
       interval: "weekly"
       day: "wednesday"
+      time: "00:00"
+      timezone: "Europe/Berlin"
     labels:
       - "Maintenance"
     open-pull-requests-limit: 999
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -8,7 +8,8 @@ updates:
   - package-ecosystem: "npm" # See documentation for possible values
     directory: "/" # Location of package manifests
     schedule:
-      interval: "daily"
+      interval: "weekly"
+      day: "wednesday"
```
nsunami, line 12: According to the documentation, it looks like the job runs on a random time when the time is not specified. 

Would it be better if we specify the time so that the dependabot PRs are ready at the start of the day? Something like below? (Trying to set that the job runs at the 00:00 AM on Wednesday) 

```suggestion
      interval: "weekly"
      day: "wednesday"
      time: "00:00"
      timezone: "Europe/Berlin"
```

nsunami (changes requested):
@chartgerink Thanks, Chris! I wanted to ask one question about setting the time of the job as below ✌️

nsunami (approved):
LGTM! 🙌

PR #952, opened 2022-12-01
Fixing the issue of sending out approval emails with wrong approver's names

### LINKED ISSUE #730: Approval email with the wrong name 🐛
### URL of the Page

_No response_

### What happened?

I received an email about an author approving a module, but it was in my name (where I'd previously approved that module).

It seems like something is going wrong in terms of the email sending. We should double check and test multi-author situations and the email sending a bit more.

### Steps to reproduce

_No response_

_No response_

nsunami:
This PR addresses the issue where the app sends out an approval email for all users, even though only one author has approved the module for publication. 

The problem was that the mailer function was sending out an approval email for all authors with their own respective name.  (e.g., Author A gets an email saying that Author A approved the module, and Author B gets that Author B approved that module, even though in reality, Author C was the one who approved the module).

I tried to address this issue by passing the `user` object from the front end to get who is approving. Then, I use that information to send all authors that this particular person has approved the module. 

Hope this works!

Fixes #730

### DIFF
--- a/app/authorship/mutations/approveAuthorship.ts
+++ b/app/authorship/mutations/approveAuthorship.ts
@@ -2,7 +2,7 @@ import { resolver } from "@blitzjs/rpc"
 import db from "db"
 import approvalMailer from "pages/api/approval-mailer"
 
-export default resolver.pipe(resolver.authorize(), async ({ id, suffix }) => {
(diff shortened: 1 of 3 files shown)


chartgerink (approved):
Works as promised! Thanks Nami 🙏

PR #955, opened 2022-12-02
Cleaning up the Next.js `Link` component to solve the warning message 

### LINKED ISSUE #954: Legacy Next.js `Link` component being noisy in the console on the homepage `/`
I was getting console messages on the root:

```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```

It seems like that our `Link` component has a [`legacyBehavior`](https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-tag:~:text=legacyBehavior%20%2D%20Changes%20behavior%20so%20that%20child%20must%20be%20%3Ca%3E.%20Defaults%20to%20false) & [`passHref` ](https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-tag:~:text=passHref%20%2D%20Forces%20Link%20to%20send%20the%20href%20property%20to%20its%20child.%20Defaults%20to%20false)on, and [using this combination with a React child requires `React.forwardRef()`](https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-functional-component).

But, our child component on the main page is a simple button, so I'd think we don't need `legacyBehavior` or `passHref`.

A fix would be to remove these from the `Link`.

nsunami:
This PR cleans up the `Link` component by removing unused `passHref` and `legacyBehavior` arguments so that we can remove the warning message on the homepage.

Fixes #954.

### DIFF
--- a/app/core/components/home/ApproachSection.tsx
+++ b/app/core/components/home/ApproachSection.tsx
@@ -107,7 +107,7 @@ export const ApproachSection = () => {
             <ApproachBox key={idx} data={data} className="w-full" />
           ))}
         </div>
-        <Link href="/signup" passHref legacyBehavior>
+        <Link href="/signup">
           <Button color="primary">Publish Openly with ResearchEquals</Button>
         </Link>
       </div>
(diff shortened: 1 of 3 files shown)


chartgerink (approved):
Thanks!

I noticed this when we upgraded - good to get this out of the way now ;)

PR #958, opened 2022-12-02
Fixing seeding the test db too much

### LINKED ISSUE #892: Limiting Faker seeding when the test DB has already been seeded
Currently, seeding happens regardless of the state of the test DB. This has created an issue where the seeding process fails with duplicate warnings, since the emails (etc) created by Faker are duplicated.

We have been resolving this issue by manually resetting the test DB. However, this method is not sustainable. Thus, we want to explore another solution.

One possible solution is to avoid running the Faker script when the DB is already seeded.

nsunami:
This PR fixes the issue where we can seed the test db too much, potentially reading to an issue of having duplicate data for the fields with unique constraints. 

I added `console.log` to output messages when the seeding script is skipping the seeding.  

Fixes #892

### DIFF
--- a/db/seeds.ts
+++ b/db/seeds.ts
@@ -487,48 +487,58 @@ const seed = async () => {
       },
     })
 
-    let user
-    for (let index = 0; index < 50; index++) {
-      user = await db.user.create({
-        data: {
-          email: faker.internet.email(),
-          role: "CUSTOMER",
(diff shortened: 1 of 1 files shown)


chartgerink (approved):
This is perfect, thanks so much!

100% a DX improvement 😄

PR #975, opened 2022-12-07
Add `Gamebook` module type

chartgerink:
This PR adds a [`Gamebook` as module type.](https://www.wikidata.org/wiki/Q642946) This is also known as a choose your own adventure game.

General note: We're starting to have a bunch of module types, it might help if we think about making it searchable instead of a dropdown?

### DIFF
--- a/db/seeds.ts
+++ b/db/seeds.ts
@@ -51,6 +51,7 @@ const seed = async () => {
     { wikidata: "Q59094171", name: "Ethical review", schema: "CreativeWork" },
     { wikidata: "Q580922", name: "Preprint", schema: "CreativeWork" },
     { wikidata: "Q1744627", name: "Classification algorithm", schema: "CreativeWork" },
+    { wikidata: "Q642946", ™name: "Gamebook", schema: "CreativeWork" },
   ]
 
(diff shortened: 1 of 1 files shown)


db/seeds.ts
```diff
@@ -51,6 +51,7 @@ const seed = async () => {
     { wikidata: "Q59094171", name: "Ethical review", schema: "CreativeWork" },
     { wikidata: "Q580922", name: "Preprint", schema: "CreativeWork" },
     { wikidata: "Q1744627", name: "Classification algorithm", schema: "CreativeWork" },
+    { wikidata: "Q642946", ™name: "Gamebook", schema: "CreativeWork" },
   ]
```
nsunami, line 55:
```suggestion
    { wikidata: "Q642946", name: "Gamebook", schema: "CreativeWork" },
```

nsunami (changes requested):
I removed the `™` character that was inserted there—otherwise looking great!

db/seeds.ts
chartgerink: That is a good catch! Very odd that that got in there 😅

PR #976, opened 2022-12-09
Fixing the problem where the cursor appears to jump in the profile edit modal

### LINKED ISSUE #858: Bio editing cursor jumping 🐛
### URL of the Page

_No response_

### What happened?

I received a report of odd cursor jumping happening for a user in the workspace bio - would be good to investigate and fix.

https://user-images.githubusercontent.com/2946344/197764171-e798cb0b-3166-4e2b-9170-08c19d7cc2e1.mov

### Steps to reproduce

_No response_

Chrome

### Anything else?

_No response_

nsunami:
This PR fixes the problem where the cursor in the profile modal appears to jump around. The cause of the issue seems to be that each keystroke was initiating re-rendering of the value, upon which the cursor would jump to the end of the text. Fixes #858 

I also noticed the same issues for other fields, and thus applied the same fix:

- First name
- Last name
- pronouns
- URL


chartgerink (approved):
Wow that the fix could be that simple!

I copied the setup of the form from elsewhere, didn't realize it was creating this issue.

Thanks @nsunami 🙏

PR #995, opened 2022-12-14
Update RDS config

chartgerink:
This imports the RDS config into flightcontrol. It's already in Flightcontrol so this is just to ensure it is consistent 😊 

Screenshot 2022-12-14 at 09 32 09

The database itself hasn't changed and I'm requesting review simply to share in the event that this is happening 😉


nsunami (approved):
Noted!

PR #1069, opened 2023-01-18
Update CoC members

chartgerink:
- Update COC members for 2023
- Update links


InquisitiveVi (approved):
I approve the members details to be removed for those members stepping down. Eventually we will archive this information for institutional memory and to also seek external inputs from former members in the future.

PR #1071, opened 2023-01-19
Specify zod schema for the fields

### LINKED ISSUE #977: Simplifying the form validation in the profile setting
While working on #976, I noticed that the form validation in the profile setting could be simplified, and thus we could possibly improve the code readability there.

I think this is a low-priority issue, but I don't want to lose it. So, I'm creating an issue and a corresponding branch. 👍

chartgerink:
It looks like all the work I can find real quick for #977 was already done.

@nsunami could you confirm this is the case? Your issue was not clear on what the exact details were so I cannot assess whether this is what you intended or too little. 😊 

Fixes #977.


nsunami:
This is 1/2 of what I was thinking 👍 

The other half is to remove the try-catch part. 

For example, with the new change, we no longer need:

```
        if (values.firstName !== workspace.firstName) {
          try {
            z.string().parse(values.firstName)
            await toast.promise(
              changeFirstNameMutation({
                firstName: values.firstName,
              }),
              {
                loading: "Saving...",
                success: "Updated first name",
                error: "Hmm that first name didn't work...",
              }
            )
          } catch (error) {
            toast.error("First name needs to be a string")
          }
        }

```

chartgerink:
Hm okay - I tried testing it a bit more and it causes a lot of hangs. Given that I did not create this issue and there seem to be additional issues, I'm going to close this and leave it for a more considerate rework of the settings portal. It looks like it is causing unnecessary issues right now, so there might've been a reason (I forgot) about why the validation works this way right now.

PR #1073, opened 2023-01-19
Allow coupons in collection checkout

chartgerink:
This PR enables coupons for checkout, finally!

It was much easier than expected. :-) The only thing left is actually creating the coupons.


nsunami:
Nice! Is there a test coupon code that I can use for testing?

chartgerink:
Fair point!

Here's a video for a test coupon `COLLECTIONS50` (50% off):

https://user-images.githubusercontent.com/2946344/213997778-da5ba183-62cd-4a23-809f-a434c01db17a.mov

For anyone reading along with us - the coupon does not work in production 😉

PR #1113, opened 2023-02-13
Add supporting membership integration

chartgerink:
After General Assembly 9, we collectively decided that the supporting memberships would become integrated into ResearchEquals. This allows for easier sign up, cancelling, and administration. 

This is a draft PR, as there's quite some work left. 
The aim is to be able to merge this on March 1st.

## Todo

- [x] Update terms (working on this with lawyer)
- [x] More...


chartgerink:
This PR is ready for review. I have not documented all too well what goes in here, but I've tested everything many many times. There will invariably still be issues to resolve, but this is a great start I think.

Notes:
* Added `/supporting-member` route for information
* Added `/membership-area` and `/membership-area/[slug]` pages
* Redid the admin panel completely
* The admin panel can be used to send emails to supporting members and add bits to the membership area itself.
* There's a success page after signing up for a supporting membership.

We will need to update the session role properly, in the near future, such that people don't need to log out and back in to be able to access the membership area.

PR #1150, opened 2023-02-27
Add ROR integration.

chartgerink:
This PR adds ROR integration. This is currently a work in progress, over the next three weeks.

Each session is livestreamed on YouTube - we hope to get this done by 20.03!

View [miro board](https://miro.com/app/board/uXjVPjRIcP8=/?share_link_id=350005303867) with notes.

## Notes ⏰ 

- [x] The mgirations for the database need to be redone after merging #1113 - otherwise we may get conflicting migrations


app/core/crossref/contributors.tsx
```diff
@@ -1,23 +1,86 @@
 import { Element, Text } from "xast"
```
tefkah, line 2: This file is getting a little messy, i could also move these all to a separate "types" file, just so they're out of the way.
Btw I took most of these from the `xsd-to-xast` generator

PR #1151, opened 2023-02-28
Update CollectionsSection.tsx

chartgerink:
Add an id to the collections section to make it easier to link to (i.e., https://researchequals.com/#collections).


chartgerink:
This doesn't work as of right now because of the way the page loads. Merging for completeness anyway.

PR #1189, opened 2023-03-10
Make the title of log in page to `R= Log In`

### LINKED ISSUE #1187: Feature request: add ResearchEquals to log in page
The current title of the login page is "Log in". It might be me not using a tool correctly, but my password manager can not perfectly match the form because there are other websites that use the same page title.

Something like "Log in | ResearchEquals" would allow a perfect match and improve my login experience :smile:

nsunami:
This PR makes the title of the log-in page to `R= Log In` from `Log In`. I chose `R= ` instead of other options since it's consistent with the titles with other pages (e.g., modules, terms). 

Fixes #1187

### DIFF
--- a/pages/login.tsx
+++ b/pages/login.tsx
@@ -19,6 +19,6 @@ const LoginPage: BlitzPage = () => {
 }
 
 LoginPage.redirectAuthenticatedTo = "/dashboard"
-LoginPage.getLayout = (page) => <Layout title="Log In">{page}</Layout>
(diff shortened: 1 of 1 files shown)


chartgerink (approved):
I'm not sure whether this will fix the issue so I'll leave the PR open for a few more days (at the latest Thursday) to see whether Daniel gets back to us. It's hard to test without the exact password manager.

I have not had issues like this before so it may be a quirk.

PR #1318, opened 2023-05-04
Add support for sponsored supporting members

chartgerink:
This PR adds all the required payment infrastructure (I think) for sponsored supporting members to sign up.

* Signup modal now takes a `sponsored` argument, which defaults to false. If `true`, it will set prices to zero and go to a new checkout session that auto-generates a discount code and applies it to the subscription.
* The signup modal for sponsored memberships only allows yearly memberships - not monthly (institutions sponsor in years)
* The `stripe_webhook` now auto-cancels the subscription after it's created. Sponsorships are on a yearly basis, so after a year the sponsorship ends. A user is always able to resubscribe with a sponsorship, if these are available when they need to resubscribe

Here's a video showing how it looks after `sponsored = true`.

https://user-images.githubusercontent.com/2946344/236178734-509048be-ad25-4ff4-8efe-a9e1c2109b5b.mov

For review, please note there is currently no implementation of the sponsored memberships. This video was created for demo purposes.

### DIFF
--- a/app/core/components/OnboardingQuests.jsx
+++ b/app/core/components/OnboardingQuests.jsx
@@ -594,7 +594,6 @@ const OnboardingSupporting = ({ data }) => {
                     Subscribe<span aria-hidden="true">&rarr;</span>
                   </span>
                 }
-                sponsored={true}
               />
             </div>
           </div>
--- a/app/core/components/OnboardingQuests.jsx
+++ b/app/core/components/OnboardingQuests.jsx
@@ -84,7 +84,7 @@ const OnboardingEmail = ({ data }) => {
             </div>
           </div>
           <div className="block text-right text-amber-700 dark:text-amber-200">
-            <p className="mt-3 text-sm md:mt-0 md:ml-6">
+            <p className="mt-3 text-sm md:ml-6 md:mt-0">
               {isSuccess ? (
                 <p className="whitespace-nowrap font-medium  underline">Email sent!</p>
               ) : (
@@ -145,7 +145,7 @@ const OnboardingEmailAccept = ({ data }) => {
             </div>
           </div>
           <div className="block text-right text-cyan-700 dark:text-cyan-200">
-            <p className="mt-3 text-sm md:mt-0 md:ml-6">
+            <p className="mt-3 text-sm md:ml-6 md:mt-0">
               {isSuccess ? (
                 <button
                   className="whitespace-nowrap font-medium underline hover:cursor-pointer hover:text-blue-600"
@@ -225,7 +225,7 @@ const OnboardingOrcid = ({ data }) => {
             </div>
           </div>
           <div className="block text-right text-emerald-700 dark:text-emerald-200">
-            <p className="mt-3 text-sm md:mt-0 md:ml-6">
+            <p className="mt-3 text-sm md:ml-6 md:mt-0">
               <a href="/api/auth/orcid" className="whitespace-nowrap font-medium  underline">
                 Connect now <span aria-hidden="true">&rarr;</span>
               </a>
@@ -280,7 +280,7 @@ const OnboardingProfile = ({ data }) => {
           </div>
           <div className="block text-right text-pink-700 dark:text-pink-200">
             <button
-              className="mt-3 whitespace-nowrap text-sm font-medium underline hover:text-blue-600 md:mt-0 md:ml-6"
+              className="mt-3 whitespace-nowrap text-sm font-medium underline hover:text-blue-600 md:ml-6 md:mt-0"
               onClick={() => {
                 setSettingsModal(!settingsModal)
               }}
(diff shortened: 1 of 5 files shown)


app/core/modals/SupportingMemberSignupModal.tsx
nsunami, line 1: Would it be a good idea to tell the user explicitly about why they are seeing the price of 0? (i.e. because they are sponsors?)

app/core/components/OnboardingQuests.jsx
```diff
@@ -587,13 +587,14 @@ const OnboardingSupporting = ({ data }) => {
...
           <div className="block text-right text-green-700 dark:text-green-200">
-            <div className="mt-3 text-sm md:mt-0 md:ml-6">
+            <div className="mt-3 text-sm md:ml-6 md:mt-0">
               <SupportingMemberSignupModal
                 button={
                   <span className="underline">
                     Subscribe<span aria-hidden="true">&rarr;</span>
                   </span>
                 }
+                sponsored={true}
```
nsunami, line 597: Just want to double-check: Is this `true` for the testing purpose? (Otherwise, I'm guessing everyone seeing this component will get free supporting membership, even though they are not sponsors.)

app/core/modals/SupportingMemberSignupModal.tsx
```diff
@@ -12,7 +12,7 @@ function classNames(...classes) {
   return classes.filter(Boolean).join(" ")
 }
 
-export default function SignupModal({ button }) {
+export default function SignupModal({ button, sponsored = false }) {
```
nsunami, line 15: The name of the component here is `SignupModal`, but the filename here is `SupportingMemberSignupModal`. Is this intended? (I'm assuming that it's better to keep these two the same.)

app/core/modals/SupportingMemberSignupModal.tsx
```diff
@@ -168,23 +172,23 @@ export default function SignupModal({ button }) {
                       type="button"
                       onClick={() =>
                         handlePayment(
-                          `/api/checkout_sessions_supporting?email=${encodeURIComponent(
-                            currentUser?.email!
-                          )}&price_id=${price}`,
+                          `/api/checkout_sessions_supporting${
+                            sponsored ? "_sponsored" : ""
+                          }?email=${encodeURIComponent(currentUser?.email!)}&price_id=${price}`,
```
nsunami, line 177: This may be an out-of-scope in this PR, but I'd think it would be a good practice to work on types so that we can avoid using the bang (`!`).

pages/api/checkout_sessions_supporting_sponsored.ts
```diff
@@ -0,0 +1,56 @@
...
+const stripe = require("stripe")(process.env.STRIPE_SECRET_KEY, { apiVersion: "2020-03-02" })
+
+const CreateSessionSupporting = async (req: NextApiRequest, res: NextApiResponse) => {
+  if (!req.query.email || !req.query.price_id) {
+    res.status(500).end("Incomplete request")
+  } else {
+    if (req.method === "POST" && req.query.price_id && req.query.email) {
+      try {
+        const coupon = await stripe.coupons.create({
+          duration: "forever",
```
nsunami, line 12: I want to double-check: Does the coupon expire as soon as it's used? "Forever" does not mean that it can be used repeatedly?

pages/api/stripe_webhook.ts
```diff
@@ -169,6 +169,11 @@ const webhook = async (req: NextApiRequest, res: NextApiResponse) => {
         },
       })
       await supportingSignup(event.data.object.customer_email)
+      // If it's a sponsored supporter, make sure to cancel the subscription immediately
+      // Sponsored memberships last only for a year
+      if (event.data.object.lines.data[0].metadata.subType === "sponsored") {
```
nsunami, line 174: A very tiny remark: this line does a lot of digging into property. I wondered if it would be better to abstract a bit for readability. For example, we could construct `isSponsored` variable to use in the if condition to improve readability for the if condition. Just a thought 👋

nsunami (changes requested):
Hi Chris! Great improvements 👍 I added quick comments in my review. They are mostly clarifications. If you want me to dive deeper, please let me know.

Side note: It took a moment for me to realize that we are taking two different things by saying "supporting member" vs. "sponsor". They both sounded similar to me at first. (Now it's all good.)

app/core/components/OnboardingQuests.jsx
chartgerink: That's an oopsie! 😅

app/core/modals/SupportingMemberSignupModal.tsx
chartgerink: Yes, great suggestion! Will add.

app/core/modals/SupportingMemberSignupModal.tsx
chartgerink: That should be updated indeed, bad habit of mine.

nsunami (approved):
These look good to me!

PR #1332, opened 2023-05-15
Add role update upon supporting

chartgerink:
This PR adds a session update when a member completes their signup process.

This removes the need for logging out and back in. Note: We still need to downgrade the role automatically when their supporting membership ends.


chartgerink:
Did a self-review and it worked 👍

PR #1381, opened 2023-06-08
Update pronouns

chartgerink:
@InquisitiveVi - inviting you to review simply because this is a change to the code of conduct, technically, and I want to encourage good process here 😊


InquisitiveVi (approved):
Change of pronouns respected and approved as an edit in the COC.

PR #1382, opened 2023-06-08
Fix `undefined undefined` in emails

chartgerink:
There was a common issue of people receiving approval emails for `undefined undefined` - documentation in screenshot:

Screenshot 2023-06-08 at 11 01 21

This was originally fixed in #952, but apparently there was a remaining bug here.

Upon some digging I discovered we are using the `user.id` to match to a `workspace` - this leads to an undefined workspace and undefined names. User IDs do not necessarily match workspace ids, but our test environments led us to believe it worked because there they were.

The fix here is to pass the approver's workspace data into the approval mailer directly. This allows us to use their names for the email, without any matching needed.


nsunami (approved):
Thanks for catching the bug that I introduced 🐛 🙏
Looks good to me!

PR #1464, opened 2023-07-11
Add filter for null values

chartgerink:
This PR adds a filter for `null` values, as the weekly digest was receiving the following template data:

```
  "collections": [
    null,
    null
  ],
```

This resulted in emails being sent even if there were no drafts etc:
Screenshot 2023-07-11 at 09 32 18

As a result of the filter, the template data will be:

```
  "collections": [

  ],
```

which results in no email being sent.


chartgerink:
Tested on my local quirrel instance.

PR #1956, opened 2024-02-12
Changing the node version described in README from 16 to 18

nsunami:
Hi @chartgerink!
I noticed that the node version described in the README was 16, but the package.json specifies 18.x. So I think would it be a good idea to update this?  

(Submitting a PR off the bat since hte change is super minor)


chartgerink (approved):
Thanks @nsunami 🙏

PR #1987, opened 2024-03-01
Update from `roles` to `role`

### LINKED ISSUE #1958: Role based routing permissions not working
We use role based routing permissions in some scenario's, which help manage access to restricted pages.

Examples include the `/admin` portal or membership pages. 

Right now, people will be redirected to the dashboard page even if they have the right permissions. It seems like BlitzJS is having some issues on managing these permissions (both in production and development).

This is actively preventing us from using the admin portal, which we use to send marketing emails and the like.

chartgerink:
This PR updates the `roles` element of the user data to `role`. It appears that BlitzJS only validates against the `role` and therefore was having issues allowing people with the right permissions to access restricted pages (e.g., `/admin` but also the membership area).

Fixes #1958.

### DIFF
--- a/app/auth/mutations/signup.ts
+++ b/app/auth/mutations/signup.ts
@@ -85,11 +85,11 @@ export default resolver.pipe(
       }),
       ctx.session.$create({
         userId: user.id,
-        roles: [user.role, user.memberships[0]!.role],
+        role: user.role,
         workspaceId: user.memberships[0]!.workspaceId,
       }),
     ])
 
     return user
-  }
+  },
 )
--- a/app/auth/mutations/login.ts
+++ b/app/auth/mutations/login.ts
@@ -37,7 +37,7 @@ export default resolver.pipe(resolver.zod(Login), async ({ email, password }, ct
 
   await ctx.session.$create({
     userId: user.id,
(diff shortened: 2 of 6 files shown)


app/auth/mutations/signup.ts
```diff
@@ -85,11 +85,11 @@ export default resolver.pipe(
...
         userId: user.id,
-        roles: [user.role, user.memberships[0]!.role],
+        role: user.role,
         workspaceId: user.memberships[0]!.workspaceId,
       }),
     ])
 
     return user
-  }
+  },
```
nsunami, line 94: It'd be nice if we add to the VS Code config file as to whether we use trailing commans or not 
e.g., `"trailingComma": "none"`

nsunami (approved):
Looks good to me 👍 
I tested setting myself as a SUPERADMIN, and I was able to access the admin dashboard.  

I had one additional comment about the prettier config for disabling trailing space.

app/auth/mutations/signup.ts
chartgerink: Thanks - it seems like this snuck in here. Happy to discuss and track in an issue #1989.

PR #1988, opened 2024-03-02
Removing section grid template on wider screen

### LINKED ISSUE #1986: Visual bug Collections modal🐛
### URL of the Page

_No response_

### What happened?

Screenshot 2024-03-01 at 10 26 13

### Steps to reproduce

_No response_

Firefox

### Anything else?

_No response_

jazellemaira:
This PR fixes the visual bug reported in issue #1986 
Collections-modal-screenshot
Fixes #1986

### DIFF
--- a/app/core/modals/CollectionsModal.tsx
+++ b/app/core/modals/CollectionsModal.tsx
@@ -104,7 +104,7 @@ export default function CollectionsModal({ button, styling, user, workspace }) {
               leaveFrom="opacity-100 scale-100"
               leaveTo="opacity-0 scale-95"
             >
-              <div className="sm:min-w-120 sm:max-w-120 inline-block min-h-screen w-full transform rounded bg-transparent text-left align-middle text-gray-900 transition-all dark:border-gray-600 dark:bg-gray-900 sm:min-h-full sm:w-auto">
+              <div className="sm:min-w-120 inline-block min-h-screen w-full transform rounded bg-transparent text-left align-middle text-gray-900 transition-all dark:border-gray-600 dark:bg-gray-900 sm:min-h-full sm:w-auto">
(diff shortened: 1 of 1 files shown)


chartgerink:
Thanks @jazellemaira for proposing this change! Appreciate you taking the time to contribute here 😊 

The current changes will change the designs. Your change makes it one column, two rows. The original design is two columns, one row. This is not very clear from #1986 - sorry for the lack of context.

The fix should end up looking something like this:

Screenshot 2024-03-03 at 11 12 34

I looked into it and it seems like there is a `sm:min-w-120` that is causing the issue in a parent div. Do you want to take another pass at this?

chartgerink (changes requested):
Thanks @jazellemaira for proposing this change! Appreciate you taking the time to contribute here 😊 

The current changes will change the designs. Your change makes it one column, two rows. The original design is two columns, one row. This is not very clear from #1986 - sorry for the lack of context.

The fix should end up looking something like this:

Screenshot 2024-03-03 at 11 12 34

I looked into it and it seems like there is a `sm:min-w-120` that is causing the issue in a parent div. Do you want to take another pass at this?

jazellemaira:
Hi @chartgerink. Thank you for telling me about the design. I looked into it further and found that it is not `sm:min-w-120` that was causing the issue to the parent div but rather the 'sm:max-w-120'. I made some changes and now it's two columns, one row as per design. 

Corrected-collections-modal-screenshot

chartgerink (approved):
Perfect, thanks a lot @jazellemaira 🙏 Merging your first contribution to ResearchEquals in 3-2-1... 😄

chartgerink:
@all-contributors add @jazellemaira for code

PR #2022, opened 2024-04-20
Fixing the inability to edit the profile when duplicate settings modals are showing up

### LINKED ISSUE #1955: 🐛 Cannot edit one's profile info - the modal closing with a click
### URL of the Page

https://www.researchequals.com/nsunami

### What happened?

When I try to edit my information on the "Edit Profile" modal, the modal closes, and thus I cannot edit the profile.

https://github.com/libscie/ResearchEquals.com/assets/17035406/00175879-661b-425f-9ec2-89989f77f1d6

### Steps to reproduce

1. Go to own profile
2. Open the Edit Profile modal by clicking the Edit Profile button 
3. Click the Name field (or any other field)

Chrome

### Anything else?

_No response_

nsunami:
This PR fixes the bug where we cannot edit the profile. (fixes #1955)

I tried to fix this by removing the use of a Recoil state to handle the <SettingsModal>'s state. Doing so will remove the issue of having the duplicate modals open, which was tied to the issue.

By implementing this fix, I faced another bug, where I could no longer open the onboarding quest. (This was also handled by the Recoil state.) To fix this, I updated the <OnboardingProfile> component, and relied on the React queries. A small issue of this approach is that it takes time for the onboarding quest card to disappear after completing the information. But, I assume this is a small cost compared to not being able to edit the profile.

We could also work on hooks to evaluate whether the profile info is complete on the SettingsModal level and below, and control the OnboardingProfile card via Context/Reducer. But then, for now, I hope this fix is enough.

Hope this works! 🤞

### DIFF
--- a/app/core/components/OnboardingQuests.jsx
+++ b/app/core/components/OnboardingQuests.jsx
@@ -1,7 +1,7 @@
 import Link from "next/link"
 import { useMutation, useQuery } from "@blitzjs/rpc"
 import { Widget } from "@uploadcare/react-widget"
-import { useRef, useState } from "react"
+import { useMemo, useRef, useState } from "react"
 import { Formik, Form } from "formik"
 import {
   Building,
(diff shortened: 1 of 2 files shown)


chartgerink:
Thanks @nsunami! I will try to review it tomorrow 😊

chartgerink (approved):
looks good to me - should not have taken me this long. I am sorry that it did. Happy to approve this now <3

PR #2024, opened 2024-04-21
Setting Prettier config on trailing comma to "ES5"

nsunami:
This PR sets the trailing comma setting on Prettier to ES5.

More info on the Prettier Doc:
https://prettier.io/docs/en/options.html#trailing-commas

I just realize that the default from Prettier is "all"—do we want "all" instead? It's going to be bit weird if "all" since we will have something like the comma in function arguments, for example:
```js
console.log(
  `Skipping seeding test users since there are already ${numUsers} users in the database.`,
)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
   },
   "prettier": {
     "semi": false,
-    "printWidth": 100
+    "printWidth": 100,
+    "trailingComma": "es5"
   },
   "lint-staged": {
     "*.{js,ts,tsx}": [


chartgerink (approved):
I think this is fine for now - if we have other insights later on we can always update then. 

Did you also apply the prettier config across the files, so that it's consistent from here on out? (e.g., `npx prettier --write`). I am happy to approve regardless - I leave it up to you to (not) apply it across the files.

PR #2069, opened 2024-05-31
Optimize queries on the dashboard

### LINKED ISSUE #2068: Optimize dashboard loading
The current dashboard takes ages to load. In our workshops using ResearchEquals, this is now a bottleneck. It is priority to try and optimize the dashboard so that this is less of an option.

Thanks @JBISchmidt for reporting this.

chartgerink:
I accidentally committed everything on main on my machine, so there may be some issues with this PR 😅 

I primarily ensured all queries happen as low on the component tree as possible, and ended up being able to remove the huge query called `getDashboardData` that caused the dashboard to slow down.

I introduced a few regressions as there is no good loading state for the dashboard, but figured it is better to have a bit of layout shift to be more responsive.

## Overview of commits

- **Prune dashboard code**
- **Remove ORCID onboarding quest**
- **Put queries for ModuleBoxFeed downstream**
- **Streamline dashboard**
- **Streamline OnboardingQuests**
- **Move queries downstream for OnboardingQuests**
- **Update components to localize queries**
- **Remove `getDashboardData` (no longer needed)**
- **Update Navbar calls**
- **Remove ORCID onboarding quest**
- **Streamline OnboardingQuests**
- **Move queries downstream for OnboardingQuests**
- **Update components to localize queries**

---

Fixes #2068 2068.


nsunami (approved):
I tested running it on local and clicking around the dashboard--it's working for me.

Great to see that many things got cleaned up! 

Also, Navbar now appears to be self-contained? Perhaps, it's a good idea to move it to Layout (just an idea).

If you need me to look in to parts of the code in more detail, happy to do so as well. But it looks good to me for now. 👍

Issue #2101, opened 2024-06-19
🐛 issue with login and email verification


McHaillet:
### URL of the Page

https://researchequals.com/dashboard

### What happened?

When I login to researchequals, I get the following screen:

![image](https://github.com/libscie/ResearchEquals.com/assets/58044494/fe3f6c31-88fe-4cd9-9de2-2d62350b316c)

After I refresh the page, I am logged in and get the expected dashboard.

Additionally, the email verification is not working. After I click the link in my email, it hangs on the following screen:

![image](https://github.com/libscie/ResearchEquals.com/assets/58044494/a02bca1e-7501-41c9-963f-d043b85bf421)

The issues could very well be related.

### Steps to reproduce

I made a new account, only changed my first and last name in my profile.

### What browser(s) did you use?

Firefox

### Anything else?

I selected firefox as the browser as I tried that right now. However, earlier today I had the same issue on Chrome.

Let me know if you need more info!

chartgerink:
Welcome and thanks @McHaillet 🙏 

On the email verification, it is a known issue that it can hang and as we're working towards our v2, this is unlikely to get fixed. Nonetheless, I checked and your email is verified despite this issue. 😊 

Regarding the log in error, I am not sure what is happening. I am unable to reproduce it at this time, even with a new account. Do you have any plugins installed in both Chrome and Firefox? Sometimes they can interfere, although I expect it is more likely an issue on our end 😅

McHaillet:
Hi @chartgerink!

> On the email verification, it is a known issue that it can hang and as we're working towards our v2, this is unlikely to get fixed. Nonetheless, I checked and your email is verified despite this issue. 😊

Okay, that is good to know :+1: 

> Regarding the log in error, I am not sure what is happening. I am unable to reproduce it at this time, even with a new account. Do you have any plugins installed in both Chrome and Firefox? Sometimes they can interfere, although I expect it is more likely an issue on our end 😅

On firefox its only an adblocker extension. On Chrome I would need to check later today.

chartgerink:
I suspect it isn't the problem then - thanks for letting me know! I will try to see whether I can reproduce. It'll be hard to fix if I can't 🤔 @nsunami have you encountered this issue?

nsunami:
I tried logging in and out several times, and I got 400 once in those times (on production).

[...]

PR #2103, opened 2024-06-24
Fixing the 400 error on dashboard page

nsunami:
This is my attempt to fix the 400 error on the dashboard page. As reported in #2101, the dashboard page was throwing a 400 error. The error message varied from `d is null` to `Cannot read properties of null (reading "firstName")`. 

When looking at the Dashboard page, I noticed that the workspace information is expected to be non-null, by non-null assertion by bang (!):

```tsx
{currentWorkspace!.firstName && currentWorkspace!.lastName
? `${currentWorkspace!.firstName} ${currentWorkspace!.lastName}`
: `@${currentWorkspace!.handle}`}
```

I assume this means that when the currentWorkspace object is indeed null (when logging out), it will throw an error.  I changed this part of the code to allow nullish values:

```tsx
{currentWorkspace?.firstName && currentWorkspace?.lastName
? `${currentWorkspace.firstName} ${currentWorkspace.lastName}`
: `@${currentWorkspace?.handle}`}
```

Other line changes are due to code formatting.

@chartgerink : Could you check logging in and out several times to see if you get a 400 error on your side?

### DIFF
--- a/app/core/components/OnboardingQuests.jsx
+++ b/app/core/components/OnboardingQuests.jsx
@@ -38,19 +38,19 @@ const OnboardingQuests = ({ expire, signature }) => {
 
   return (
     <>
-      <OnboardingSupporting data={currentUser.supporting} />
-      <OnboardingEmail data={currentUser.emailIsVerified} />
-      <OnboardingEmailAccept data={currentUser} />
-      <OnboardingAvatar
-        data={currentWorkspace}
-        expire={expire}
-        signature={signature}
-      />
-      <OnboardingProfile data={currentWorkspace} />
-      <OnboardingDraft data={currentWorkspace} />
-      <OnboardingCollection data={currentWorkspace} />
-      <OnboardingAffiliation data={currentWorkspace} />
-      <OnboardingDiscord />
+      {currentUser && (
+        <>
+          <OnboardingSupporting data={currentUser.supporting} />
+          <OnboardingEmail data={currentUser.emailIsVerified} />
+          <OnboardingEmailAccept data={currentUser} />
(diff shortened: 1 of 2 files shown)


nsunami:
Update: Now I get

```
Unhandled Runtime Error
TypeError: Cannot read properties of null (reading 'supporting')
```


image

I suspect that the redirect may not be working well some way. Adding redirect may fix:

```
Dashboard.authenticate = { redirectTo: "/login" }
```

chartgerink:
Ah yes - thanks for trying to fix this 🙏 

The `currentUser` is indeed potentially not available when people directly go to the `/dashboard` route (previously `d is null` or similar). What we could do, is wrap the OnboardingQuests in a conditional render only when `currentUser` is not null. This would be in addition to your already proposed changes.

It would look something like this:

```

return (
  <>{currentUser && 
    <>
      <OnboardingQuest1 />
    </>}
  </>
)
```

I suspect this will fix both our problems, but I would encourage a bit of testing and trying to break it 😊

nsunami:
@chartgerink thanks! I adopted your suggestion. I tried logging in and out while on /dashboard multiple times, and I no longer get an error.

Could you check?

chartgerink:
@nsunami thanks for the updates and patience - I was out on vacation 🌴 

I think this is sufficient for now - I do get a bit of a flashy screen that I would personally consider problematic for production in general. 👇 (it's worse than this at times)

https://github.com/user-attachments/assets/e4eb5639-1049-4f70-a6b3-97fa5d309349

However, given that we are working towards a v2 and this is such a specific case, I think this is acceptable in the scenario of keeping the current version functional until v2 launches.

PR #2134, opened 2024-07-19
Delete .github/dependabot.yml

chartgerink:
Given that this version of ResearchEquals is on life support until the version 2 is launched, I propose we remove the dependabot upgrades to reduce the noise on the repo. 

@nsunami - what do you think?


nsunami (approved):
@chartgerink That's ok with me 👍

PR #2153, opened 2025-03-03
Remove plausible

chartgerink:
We're not counting through Plausible until 2.0. Historic data is available in our archives.


nsunami (approved):
Looks good to me!

PR #2156, opened 2025-03-16
Remove support chat

chartgerink:
As we work towards 2.0, we are reducing some of the overhead on the production version. The chat is barely used, and this allows us to keep track of one fewer things.

This does not mean we do not provide any support, only that we do not have a chat bubble at this time.

### DIFF
--- a/app/core/components/InstitutionalSupportingMembers.tsx
+++ b/app/core/components/InstitutionalSupportingMembers.tsx
@@ -1,4 +1,3 @@
-import { Crisp } from "crisp-sdk-web"
 import Tiu from "../icons/tiu"
 import KuLeuven from "public/images/kuleuven.png"
 import Oscl from "public/images/oscl.png"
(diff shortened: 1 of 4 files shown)


nsunami (approved):
Looks good to me 👍 Merging ✅

nsunami (changes requested):
@chartgerink Hi Chris! I noticed that the package lockfile had 600 lines of addition. I expected that with deleting the Crips package, we will end up with a smaller lockfile. Is it an expected outcome?

chartgerink:
Thanks @nsunami - I hadn't noticed 😅  

I tried reverting this locally; it seems to be a side effect of running `npm install` or `npm remove <pkg>`. I would not consider this blocking, but maybe I am missing something. What do you think?

nsunami:
@chartgerink I agree that it's not a stopper and thus marking this as approved ✅

PR #2168, opened 2025-05-07
fix(browse-modules): Render only R= modules on the browse page

### LINKED ISSUE #2167: Links to non R= publications in `/browse`
On https://www.researchequals.com/browse there are outgoing links to non-modules. For example:

- https://www.researchequals.com/modules/osf.io%2Fjb72d_v1
- https://www.researchequals.com/modules/pnas.2401231121
- https://www.researchequals.com/modules/zenodo.15234684
- https://www.researchequals.com/modules/m9.figshare.28545578

It seems like there is a bug in the query that is not correctly filtering. @nsunami, could you take a look?

nsunami:
This PR renders only R= modules on the browse page, relying on the `publishedWhere` attribute.

Fixes #2167

### DIFF
--- a/app/modules/queries/getBrowseModuleFeed.ts
+++ b/app/modules/queries/getBrowseModuleFeed.ts
@@ -26,6 +26,7 @@ export default resolver.pipe(
           ...paginateArgs,
           where: {
             published: true,
+            publishedWhere: "ResearchEquals",
           },
           orderBy: [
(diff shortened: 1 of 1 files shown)


chartgerink (approved):

chartgerink:
Thanks @nsunami 🙏

PR #2171, opened 2025-05-26
Fix parsing of DataCite API attributes

chartgerink:
The current implementation erroneously defaults to the publication year, and as a result, all DataCite publications will be parsed as being on the same date of the correct publication year. This is incorrect, as reported by @cmeesters.

This fix ensures that the registered or creation date are used, with registered taking precedence (equates published).


nsunami (approved):
It looks good to me!

